Issue #1, opened 2018-04-01
Provide configuration for Kubernetes


alexellis:
The public and development version of OpenFaaS Cloud is currently hosted on Docker Swarm, but we should add a configuration for Kubernetes so that we can take advantage of network-policy and other features.

* Kubernetes-style YAML files in a `yaml` folder (not helm)
* Update any references to "gateway:8080" to an environmental variable i.e. `gateway_url` or something else if it's already been used in stack.yml

Notes:

* Buildkit used in of-builder is a privileged container, so set that securityContext too

alexellis:
@viveksyngh Hi, thanks for doing some work on this. I merged a PR from @johnmccabe and now we have an additional service for buildkit. PTAL and update the YAML files.

I've also provided a way to configure the registry_url/prefix.

PR #9, opened 2018-04-04
Provide kubernetes configuration

viveksyngh:
This changes provider kubernetes configuration for of-builder and
registry as .yml file in yaml folder.

Also replaces hardcoded gateway URL with environment variable
gateway_url.

Fixes: #1

### DIFF
--- a/yaml/of-buildkit.yml
+++ b/yaml/of-buildkit.yml
@@ -1,0 +1,38 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: of-buildkit
+  namespace: openfaas
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: of-buildkit
+    spec:
+      containers:
+      - name: of-buildkit
+        args: ["--addr", "tcp://0.0.0.0:1234"]
+        image: alexellis2/buildkit:2018-04-17
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 1234
+          protocol: TCP
+        securityContext:
+          privileged: true
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: of-buildkit
+  namespace: openfaas
+  labels:
+    app: of-buildkit
+spec:
+  type: ClusterIP
+  ports:
+    - port: 1234
+      protocol: TCP
+      targetPort: 1234
+  selector:
+    app: of-buildkit
--- a/yaml/of_builder.yml
+++ b/yaml/of_builder.yml
@@ -1,0 +1,37 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: of-builder
+  namespace: openfaas
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: of-builder
+    spec:
+      containers:
+      - name: of-builder
+        image: openfaas/of-builder:0.3.0
+        imagePullPolicy: Always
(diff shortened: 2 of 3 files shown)


alexellis:
Hi Vivek, thanks for raising this PR. Have you been able to test the changes too?

CoderAway28th:
The namespaces are unmatched. @viveksyngh . Of_builder and registry namespace is "openfaas"
while gh-push and git-tar .. namespace is "openfaas-fn". That may cause nslookup error if you don't specify namespace for of_builder and registry urls.

alexellis:
Can we change point 1 so that it uses a unix timestamp instead?

viveksyngh:
@alexellis This is ready to merge along with #25 . Both has been tested on Docker swarm and k8s.

alexellis:
OK this is on my list to review.. if anyone else wants to try it out that'd help too.

alexellis:
cc @affix @johnmccabe

affix:
@alexellis @johnmccabe @viveksyngh do we really need to expose these services to the network?

If they are all called from inside the k8s namespace and are not supposed to have external comms then we shouldn't publish the services.

A ReplicaSet instead of a Service may be more suited to the builders as it makes it easier to scale in and out without exposing the services publically.

yaml/of_builder.yml
```diff
@@ -0,0 +1,38 @@
...
+  namespace: openfaas
+  labels:
+    app: of-builder
+spec:
+  type: NodePort
+  ports:
+    - port: 8080
+      protocol: TCP
+      targetPort: 8080
+      nodePort: 31121
```
alexellis: The nodePort can be removed, it'd only be needed for testing the of-builder manually. Maybe we can use a [ClusterIP](https://kubernetes.io/docs/concepts/services-networking/service/) instead?

yaml/registry.yml
```diff
@@ -0,0 +1,36 @@
...
+  namespace: openfaas
+  labels:
+    app: registry
+spec:
+  type: NodePort
+  ports:
+    - port: 5000
+      protocol: TCP
+      targetPort: 5000
+      nodePort: 31115
```
alexellis: @affix I think we could push from the of-builder/of-buildkit service but not pull from it - the Kubelet wouldn't deploy from a ClusterIP

yaml/of-buildkit.yml
```diff
@@ -0,0 +1,38 @@
...
+          privileged: true
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: of-buildkit
+  namespace: openfaas
+  labels:
+    app: of-buildkit
+spec:
```
alexellis: This NodePort can be removed in favour of a ClusterIP. I've also updated the README file for Swarm.

alexellis:
Let's merge this and then have someone with Kubernetes experience give it a once-over. Thanks for getting us this far @viveksyngh 👍

affix:
i'll deploy it tonight @alexellis (will stick it on the same cluster as my normal faas deployment in a different namespace)

viveksyngh:
@alexellis I had updated this to used `ClusterIP` instead of `NodePort`.  @affix Let me know if you face any issue in running this. There are few things that need to be configured in a different way for kubernetes, which I have listed in previous comments.

alexellis:
Lots of config changes are needed in the yaml files where we refer to `gateway` it must be `gateway.openfaas` etc

PR #11, opened 2018-04-08
miss '\' in deploy scripts

lmxia:
miss '\' at the end of second line


alexellis:
I've added this change via a related update. Thank you for raising it.

PR #14, opened 2018-04-16
a valid label must consist of alphanumeric characters, '-', '_' or '.…

lmxia:
A valid label must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character. time.Now().String() contain unaccepted characters.


viveksyngh:
I think we can use this formatting option `2006-01-02-T15-04-05Z07-00` which will have time information as well as time zone.

alexellis:
@viveksyngh where does this time format come from? Does it have a name like ISO ..xyz.. ?

alexellis:
Fixed via #24 @viveksyngh

PR #18, opened 2018-04-17
Add audit function for gh-push

alexellis:
Adds auditing for gh-push events so that we can get visibility
of what's going on in the system.

New audit function created for logging but this could be anything
even an external web service.

## Testing

Sent an event without giving `X-Github-Event` header and then giving a bad value (not "push") and then giving it again with "push"

In each instance I checked out the logs for the audit function to see messages being sent from the function.


alexellis:
@ericstoekl @johnmccabe

PR #25, opened 2018-04-25
Allow port as part of repository_url env variable

viveksyngh:
This changes to allow port as part of repository_url env variable which
will allow us to specify ports other than 5000 for registry. This will
be required for kubernetes.

Tested on local for kubernetes. Needs to be tested on Swarm.

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -69,7 +69,7 @@ func Handle(req []byte) string {
 		gatewayURL := os.Getenv("gateway_url")
 
 		// Replace image name for "localhost" for deployment
-		imageName = repositoryURL + imageName[strings.Index(imageName, ":"):]
+		imageName = getImageName(repositoryURL, imageName)
 
 		serviceValue = fmt.Sprintf("%s-%s", event.owner, event.service)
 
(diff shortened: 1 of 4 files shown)


viveksyngh:
I need to update the `gateway_config.yml` file to pass the port with repository URL. I have to also test this on docker swarm.

alexellis:
OK. I'll mark as WIP - let me know when it's fully tested

buildshiprun/handler.go
```diff
@@ -59,7 +59,7 @@ func Handle(req []byte) string {
 		gatewayURL := os.Getenv("gateway_url")
 
 		// Replace image name for "localhost" for deployment
-		imageName = repositoryURL + imageName[strings.Index(imageName, ":"):]
+		imageName = repositoryURL + imageName[strings.Index(imageName, "/"):]
```
alexellis: Please break it out into a function with a unit test

viveksyngh: Ok

viveksyngh:
I have added that as a function and tests.

alexellis:
There's another place where we have this hard-coded. Please see ops.go in git-tar - this should also be configured / changed.

viveksyngh:
in `ops.go` should not we use another variable `repository_push_url` as we discussed in the slack group?

alexellis:
@viveksyngh thanks for working on this 👍 it looks good now, but needs a rebase before I can merge it into Git.

Can you also confirm if you have tested the change end to end on Kubernetes?

viveksyngh:
I have update this PR. Yes, I have tested it end to end on kubernetes.

alexellis:
Thank you for doing that @viveksyngh - I'm going to merge now and check it out locally.

Alex

PR #26, opened 2018-04-26
Add git commit SHA to function labels

viveksyngh:
This changes adds a new label Git-SHA to functions deployed by openfaas
cloud.

Tested on local on kubernetes.

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -45,6 +45,7 @@ func Handle(req []byte) string {
 		service := os.Getenv("Http_Service")
 		owner := os.Getenv("Http_Owner")
 		repo := os.Getenv("Http_Repo")
+		sha := os.Getenv("Http_Sha")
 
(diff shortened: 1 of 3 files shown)


buildshiprun/handler.go
```diff
@@ -67,6 +67,7 @@ func Handle(req []byte) string {
 				"Git-Owner":      owner,
 				"Git-Repo":       repo,
 				"Git-DeployTime": strconv.FormatInt(time.Now().Unix(), 10), //Unix Epoch string
+				"Git-SHA":        getGitSHA(imageName),
```
alexellis: I was hoping we would pass it in from the previous method.. are we able to do that?

viveksyngh: oh yeah .. we can do that too. I think that would be a better way to do it.

alexellis:
LGTM

alexellis:
Tested - works fine

PR #30, opened 2018-04-29
Add github commit pending status

s8sg:
these changes add pending github commit status for openfaas-cloud
- add status report code in sdk
- vendor updated sdk in gh-push, git-tar and buildshiprun
- report status pending on accepted build in gh-push
- report failure on accepted build in gh-push and git-tar


s8sg:
- We are authorizing for status report everytime, we can save the token in the sdk status report library
- Also we can forward the token in between functions

I’m waiting for comments on the same

alexellis:
I see two options here:

- move the status update code into the SDK library
- create a single function which can be called to update the statuses

I think you've gone for the first option, but wdyt?

alexellis:
I see a commit that says: "Store auth token locally"

Please can you explain how this is being stored? There is no shared memory between requests (so that wouldn't work) and we shouldn't store it on the disk because it could expire between requests.

The only way I could think this would work with the classic watchdog is if it was kept higher up in the call stack when iterating over several functions.

s8sg:
I have moved the status update code into the SDK library as it is a common functionality and also the status format should be predefined, like having const for status : 
```go
const (
    PENDING = "pending",
    SUCCESS = "success",
    FAILURE = "failure"
    BUILD   = "build"
    DEPLOY = "deploy"
)
```
also the EventInfo structure is reduced the no of arguments and needs to be defined in each functions if we don't use sdk.

PR #33, opened 2018-04-29
Add pending status for service

s8sg:
This commit adds pending status for service. In a successful deployment
of all functions to buildshiprun, the deployment is marked as successful
- vendor derek at git-tar
- use <public_url>/overview?user=<>&sha=<>&repo=<> as successful commit
status url


git-tar/function/ops.go
```diff
@@ -29,6 +33,21 @@ type cfg struct {
 	Frontend *string `json:"frontend,omitempty"`
 }
 
+type eventInfo struct {
```
alexellis: is this duplicated from buildshiprun?

git-tar/function/ops.go
```diff
@@ -181,6 +200,14 @@ func deploy(tars []tarEntry, pushEvent sdk.PushEvent, stack *stack.Services) err
 	c := http.Client{}
 	gatewayURL := os.Getenv("gateway_url")
 
+	event := getEvent(pushEvent)
+	serviceValue := owner + "-" + repoName
+
+	reportStatus("pending", fmt.Sprintf("%s build and deploy is in progress", serviceValue), "DEPLOY", event)
```
alexellis: Is `DEPLOY` our own term or a GitHub constant? Could it be `Deploy`?

git-tar/function/ops.go
```diff
@@ -209,9 +236,100 @@ func deploy(tars []tarEntry, pushEvent sdk.PushEvent, stack *stack.Services) err
...
+			failedFunctions[tarEntry.functionName] = reqErr.Error()
 		}
 
 		fmt.Println("Service deployed ", tarEntry.functionName, res.Status, owner)
 	}
+
+	if !hasFailure {
+		reportStatus("success", fmt.Sprintf("service successfully deployed as: %s", serviceValue), "DEPLOY", event)
+	} else {
+		reportStatus("failure", fmt.Sprintf("failed to deploy service %s: %v", serviceValue, failedFunctions), "DEPLOY", event)
```
alexellis: We can't report status on a single service/function - we have to do it for the whole stack as a yes/no. Why? Because what if one passes then the following fails and the one after that passes? The final status will say passed and hide the error..

git-tar/function/ops.go
s8sg: Yes, but I haven’t put anything in sdk. Tried to make it as simple as possible

git-tar/function/ops.go
s8sg: Its a status context and our own term. We can definitely set it as Deploy

git-tar/function/ops.go
s8sg: As I understand if the stack has one or more failure the status should be a failure. Currently `hasFailure` is `false` by default, it's being set to `true` only when a function fails, and then there is no way it can be set to `false`.  That means if there is one function failure the whole stack would be `failure`, which I think is intended. 

I have added a `failedFunctions map[string]string` which would map `functionname` -> `error` for failed functions

git-tar/function/ops.go
s8sg: It also going to break `of_badge_gen` as the image are mapped by `status + context` [for example:  `success_BUILD.svg`], is it okay to update it in a separate PR?

s8sg:
implemented by #43

PR #37, opened 2018-04-30
Add build status as failure in case of code panic

s8sg:
This commit adds a defer panic code to update failure status in case
there is some Panic


buildshiprun/handler.go
```diff
@@ -33,6 +33,14 @@ func Handle(req []byte) string {
 		log.Panic(eventErr)
 	}
 
+	defer func() {
```
alexellis: This is an interesting approach - I need to look into panic recovery in Go and comment back on this.

Issue #2, opened 2018-04-01
Provide feedback via outgoing webhook


alexellis:
An outgoing webhook URL configured via environmental variable should allow for events to be posted to a remote server for status.

* Whenever a GH-push is received
* Whenever buildshiprun executes
* Whenever of-builder pushes a new image successfully or if it fails and how long it took

The payload format should be in JSON and include context:

i.e.

```json
{
 "owner": "alexellis",
 "repo": "kubecon-tester",
 "source": "gh-push"
}
```

The URL to receive web hooks should be set the stack.yml file - probably in the same file that has the gateway configuration URL.

A specific event would extend the JSON above with specific information like a "message" field:

```json
{
 "owner": "alexellis",
 "repo": "kubecon-tester",
 "source": "gh-push",
 "message": "owner not in CUSTOMERS file"
}
```

We may also want some additional fields to represent whether this is a success / warning message such as in the example above.

Since all OpenFaaS functions are written in Go we should have a common package with the struct representing the JSON event and a HTTP client/wrapper for easy auditing of events.

ericstoekl:
Derek assign: me

alexellis:
@ericstoekl did you see the changes that went up?

ericstoekl:
Haven't had time yet due to the faas-cli work, but I'll find some cycles to work on it today.

ericstoekl:
Now that the `sdk.PostAudit` function is in the SDK, the real issue here is figuring out what we want the audit-event to go to. I have been using `webhookstash` for testing, but we should have a database for this. Maybe we deploy minio with the of-cloud? Maybe we just use a simple database like sqllite for now?

alexellis:
Derek: close

alexellis:
It should be posted to HTTP and at that point it's agnostic as to whether you want to store that in a DB via a function, in a Git log, Kafka transaction log or a Slack channel.

PR #38, opened 2018-05-01
[Changes needed] Add audit calls to all functions

ericstoekl:
Add audit calls to all functions

Signed-off-by: Eric Stoekl <ems5311@gmail.com>

Addresses issue: https://github.com/openfaas/openfaas-cloud/issues/2

This change adds audit events to each of the functions in the project. Including the SDK in functions causes the large number off added `vendor/` files. Please let me know if this is not the correct approach.


git-tar/function/handler.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/openfaas/openfaas-cloud/sdk"
 )
 
+const Source = "gh-tar"
```
iyovcheva: IMO it's better if this is separated into different commits, keeping each one atomic and per function. Will be easier to follow the changes in the git log and easier to revert or fix if some of the functions is broken. Also better for reviewing. WDYT?

iyovcheva: As this one - [#18](https://github.com/openfaas/openfaas-cloud/pull/18)

alexellis:
It looks like vendoring the sdk package is pulling in way too much from the repo. Please could you two look into the usage of the `dep` tool and see if you can vendor just the sub-package named `sdk`?

git-tar/function/handler.go
alexellis: Yes smaller separate commits would be ideal - per function. What do you think @ericstoekl - would you have time to do that? We also need to figure out the vendoring issue also mentioned by @s8sg on Slack.

ericstoekl:
Sure, that sounds good. I agree that this pulls in way too many files. I'll work with @s8sg on vendoring the sub package.

iyovcheva:
@ericstoekl I think a `Gopkg.toml` constraint will work.

iyovcheva:
@ericstoekl addressed the `sdk` issue in [#1](https://github.com/ericstoekl/openfaas-cloud/pull/1)
Can you please review?

ericstoekl:
Thanks to @iyovcheva for the PR -- the number of files changed has been reduced by a lot as a result. :100:

iyovcheva:
Now noticed that I missed to apply the changes to `garbage-collect`.
Sorry for the noice with this additional  PR. You can squash if you prefer a clean log.
[#2](https://github.com/ericstoekl/openfaas-cloud/pull/2)

git-tar/function/handler.go
ericstoekl: @alexellis @iyovcheva Thanks for the feedback. I've separated this PR into a separate commit for each function.

git-event/handler.go
```diff
@@ -55,6 +56,14 @@ func Handle(req []byte) string {
 
 	}
 
+	auditEvent := sdk.AuditEvent{
+		Message: "Garbage Collection Completed",
+		Owner:   event.Installation.Account.Login,
```
iyovcheva: If `eventHeader ` is different than `installation_repositories`, then GC will not be completed and the message can be confusing, the empty owner as well.

If the message goes into a variable, that can solve the problem.

gh-push/Gopkg.toml
```diff
@@ -24,3 +24,7 @@
 [[constraint]]
   branch = "master"
   name = "github.com/alexellis/hmac"
+
+[prune]
```
iyovcheva: There are vendor files out of the sdk package here.

ericstoekl:
Just updated to account for @iyovcheva 's changes, also removed my changes to the `git-push` folder because there is no changes needed there. PTAL

git-event/handler.go
```diff
@@ -27,7 +29,6 @@ func Handle(req []byte) string {
 			}
 		}
 
-		event := InstallationRepositoriesEvent{}
```
iyovcheva: After auditEvent is moved to the if block, this can stay here. But this is not an issue. Feel free to ignore the comment if you don't have time to address.

iyovcheva:
Looks good

buildshiprun/handler.go
```diff
@@ -107,6 +108,15 @@ func Handle(req []byte) string {
 		log.Println(result)
 	}
 
+	auditEvent := sdk.AuditEvent{
```
alexellis: what if it's not successful?

garbage-collect/handler.go
```diff
@@ -46,6 +47,15 @@ func Handle(req []byte) string {
 		}
 	}
 
+	auditEvent := sdk.AuditEvent{
```
alexellis: Please do a rebase with the latest changes on this function. The garbage collection should post useful information such as the amount of deleted functions. See my latest update.

git-tar/function/handler.go
```diff
@@ -60,6 +62,15 @@ func Handle(req []byte) []byte {
 		log.Println(err)
 	}
 
+	auditEvent := sdk.AuditEvent{
+		Message: "Tar Completed",
```
alexellis: What other metadata do we have to add here? Number of functions for instance.

alexellis:
I can see the separate commits now, but I think what I expected was separate commits in separate PRs. I think we can work with what's been done though.

Think of the audit as a stack-trace - something which doesn't just show the start/finish of a function but any additional metadata that would help debug a problem or monitor the system. So if you have garbage collect - show how many functions are deleted (if any). If you have Git-tar - what was the filesize? How many functions were in the stack?

alexellis:
@ericstoekl I've given feedback on this PR which I think we need to tackle before merging. There's also a small rebase conflict now with the "garbage-collect" handler.

ericstoekl:
@alexellis thanks for the feedback, I'm a little tied up with work stuff tonight but I will have a resolution+update to this PR before the end of day tomorrow.

ericstoekl:
@alexellis I actually want to start submitting these as separate PRs as you have mentioned. It'll make it easier to test and iterate on each function.

PR #39, opened 2018-05-01
Fix case-sensitive username issue for router

### LINKED ISSUE #34: Issue: case-sensitive usernames are not supported by the router
The HTTPS router that gives a sub-domain to each user will not support a username like @BurtonR - burtonr would be fine though.

> An Internet address is only case sensitive for everything after the domain name. 

https://www.computerhope.com/issues/ch000709.htm

It would be simple to have the functions deployed with lowercase metadata for usernames - this would resolve the problem. The `Git-Owner` label should be updated in the `buildshiprun` function.

viveksyngh:
This fixes issue with case-sensitive username issue for the router

Fixes #34 

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -86,7 +86,7 @@ func Handle(req []byte) string {
 			Network: "func_functions",
 			Labels: map[string]string{
 				"Git-Cloud":      "1",
-				"Git-Owner":      event.owner,
+				"Git-Owner":      strings.ToLower(event.owner),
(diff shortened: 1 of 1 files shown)


alexellis:
Unfortunately I think this will break in other areas - like where we reconstruct a URL to point at GitHub. domain names are not case sensitive but the resource part of the URL is.

alexellis:
Unfortunately I think the change is being applied in the wrong place so will close this PR.

PR #44, opened 2018-05-07
of-builder: use Go client instead of buildctl binary

AkihiroSuda:
Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>


of-builder/Dockerfile
```diff
@@ -1,15 +1,15 @@
...
+FROM golang:1.10-alpine AS builder
 RUN apk add --no-cache git g++ linux-headers
 WORKDIR /go/src/github.com/alexellis/of-cloud/of-builder
 ADD main.go .
 ADD vendor  vendor
 
 RUN go build -o /usr/bin/of-builder .
 
-FROM alexellis2/buildkit:2018-04-17
+FROM alpine:3.7
```
alexellis: Nice side-benefit.

of-builder/Dockerfile
```diff
@@ -1,15 +1,15 @@
-FROM golang:1.9-alpine AS builder
+FROM golang:1.10-alpine AS builder
```
alexellis: Is this version move required for the change?

of-builder/main.go
```diff
@@ -88,17 +92,50 @@ func build(w http.ResponseWriter, r *http.Request) ([]byte, error) {
...
+			"dockerfile": contextDir,
+		},
+		Frontend: "gateway.v0",
+		FrontendAttrs: map[string]string{
+			"source": cfg.Frontend,
+		},
+		// ~/.docker/config.json could be provided as Kube or Swarm's secret
+		Session: []session.Attachable{authprovider.NewDockerAuthProvider()},
+	}
+	c, err := client.New("tcp://of-buildkit:1234", client.WithBlock())
```
alexellis: This address should probably move to an env-var for configuration. @johnmccabe how did you come to pick 1234 as a port in your original PR? (is this a convention for gRPC?)

of-builder/main.go
```diff
@@ -88,17 +92,50 @@ func build(w http.ResponseWriter, r *http.Request) ([]byte, error) {
...
+	}
+	ch := make(chan *client.SolveStatus)
+	eg, ctx := errgroup.WithContext(context.Background())
+	eg.Go(func() error {
+		return c.Solve(ctx, nil, solveOpt, ch)
+	})
+	eg.Go(func() error {
+		for s := range ch {
+			for _, v := range s.Vertexes {
+				logrus.Debugf("vertex: %s %s %v %v", v.Digest, v.Name, v.Started, v.Completed)
```
alexellis: @AkihiroSuda I've used Go's `log` in the other packages due to headaches with logrus and vendoring - after their upper/lowercase rename it made things difficult. Is logrus required?

alexellis:
It would be a breaking change - but is it now easier to return a set of build logs from the container builder?

of-builder/main.go
johnmccabe: Its buildkits default, see [here](https://github.com/moby/buildkit/tree/896e65bd0978ae3b74e78d7ae8f19ef6d906740c#running-containerized-buildkit).

of-builder/Dockerfile
AkihiroSuda: Build fails with 1.9
```
Step 6/11 : RUN go build -o /usr/bin/of-builder .
 ---> Running in b7782fb9892d
# github.com/alexellis/of-cloud/of-builder/vendor/github.com/docker/docker/pkg/archive
vendor/github.com/docker/docker/pkg/archive/archive.go:364:5: hdr.Format undefined (type *tar.Header has no field or method Format)
vendor/github.com/docker/docker/pkg/archive/archive.go:364:15: undefined: tar.FormatPAX
vendor/github.com/docker/docker/pkg/archive/archive.go:1166:7: hdr.Format undefined (type *tar.Header has no field or method Format)
vendor/github.com/docker/docker/pkg/archive/archive.go:1166:17: undefined: tar.FormatPAX
The command '/bin/sh -c go build -o /usr/bin/of-builder .' returned a non-zero code: 2
```

If 1.10 is problematic we would need to vendor older version of docker/docker (and buildkit as well maybe)

of-builder/main.go
AkihiroSuda: updated to use `log`, but `logrus` is still vendored because other vendored packages import logrus

AkihiroSuda:
> is it now easier to return a set of build logs from the container builder?

Yes, but I'd suggest streaming logs and vertex statuses (with always HTTP 200 status code) rather than returning them at once.

of-builder/Dockerfile
alexellis: This container is going to be self-contained so it may not be so much of an issue - but the package with issues in 1.10 (tar / os/user ) is used in other functions. https://github.com/golang/go/issues/24787

AkihiroSuda:
```
HTTP/1.1 200 OK
Content-Type: application/x-ndjson


{Single-line JSON representation of BuildKit SolveStatus structure which contains per-vertex statuses and logs}
{...}
...
```

PR #46, opened 2018-05-12
Fix misspelling of my Github name in CUSTOMERS

johnmccabe:
This is a trivial type fix, looks like a misspelling of my twitter handle had been used here, this PR fixes the github name for @johnmccabe.


johnmccabe:
@alexellis

alexellis:
Derek :+1:

PR #55, opened 2018-05-20
Limit number of function per repository

### LINKED ISSUE #50: Limit # of functions per repo
An exhaustion attack on OpenFaaS Cloud could involve adding hundreds of functions to a stack file to exhaust resources and cause noise.

If these functions were deployed and created Pods then they would also cause the cluster to fill up.

We should look to add a config limit for how many functions are allows per repo/stack file.

I am thinking of a default limit of 5

viveksyngh:
This changes limits number of function per repository for OpenFaaS
cloud. By default it is 5 and can be changed using
`functions_limit_count` environment variable in `gateway_config.yml`

Fixes #50

### DIFF
--- a/git-tar/function/handler.go
+++ b/git-tar/function/handler.go
@@ -46,8 +46,10 @@ func Handle(req []byte) []byte {
 		os.Exit(-1)
 	}
 
-	if ok, errMessage := isWithinLimit(len(stack.Functions)); !ok {
-		log.Println(errMessage)
+	if ok, limitErr := isWithinLimit(len(stack.Functions)); !ok {
+		log.Println("Limit error: ", limitErr.Error())
+		status.AddStatus(sdk.StatusFailure, "Limit error : "+limitErr.Error(), sdk.StackContext)
+		reportStatus(status)
 		os.Exit(-1)
 	}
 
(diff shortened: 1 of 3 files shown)


alexellis:
Hi @viveksyngh thanks for working on this. I think there is some work remaining before the merge.

I think this may belong in github.yml?

The name may be better named as `functions_per_repo_limit` to be clearer that this limit applies to the functions per repo. Then we'll only apply this setting if it is present. I'd like to see a couple of unit tests showing this flag is in action.

Don't assume the flag is always present, but is `os.LookupEnv` to check first.

Please add these tests:

- present and value is not a number
- present and value causes rejection of a stack file with too many functions
- not present and stack file still gets deployed / validated

Can you think of anything else we need to test? If you run into issues testing the whole function then create a new function that can extract this functionality. 

Alex

viveksyngh:
I have added test cases. Also, I have tested this on my local for all scenario.

git-tar/function/handler.go
```diff
@@ -36,6 +37,11 @@ func Handle(req []byte) []byte {
 		os.Exit(-1)
 	}
 
+	if ok, errMessage := isWithinLimit(len(stack.Functions)); !ok {
```
alexellis: This looks great, but we need to do a little more before releasing.

This should communicate back via audit and also via GitHub statuses. As it is now the user won't get any feedback.

viveksyngh: Ok, I will add that.

viveksyngh: @alexellis I think @s8sg has already added most of the code required to add git-status in PR #63. I think it makes more sense to make me these changes once his PR gets merged.

alexellis:
Please work together on this.

viveksyngh:
This one I tested on my local repo against https://github.com/viveksyngh/openfaas-functions set of functions. 

Status is getting updated properly in case repo exceeds the limit. 

screen shot 2018-08-17 at 6 24 38 pm

alexellis:
This is a good start 👍 We just need to think about what happens if someone deploys from multiple repos all at the same time. Technically there will be 0 functions deployed so they won't have met the limit until after the build/deploy.

Remove all integrations
Setup 2 repos with 10 functions
Do the "git push" at the exactly same time

Result: You'd get 20 functions created bypassing the limit.

burtonr:
@viveksyngh There are some conflicts on this now. Could those get resolved easily?

@alexellis do you think we could worry about that edge-case scenario on a later PR/issue? Seems like that situation would be rare, although possible. Since there hasn't been a limit yet, should we block this update for that situation? 

Seems like there would need to be some thought and discussion around that. Perhaps a new issue would make it more visible for input.

alexellis:
This didn't quite work how we hoped, but it was a useful learning experience. Thank you @viveksyngh 

I'll close.

PR #61, opened 2018-05-23
Update test cases name for GetImageName tests

### LINKED ISSUE #60: Unable to read tests for func GetImageName
Hi @viveksyngh 

Please could you refactor the names of the test cases in buildshiprun?

Currently they are named like "test case 1" and "test case 2" - these should be representative of what we're testing in each scenario. 

Thanks,

Alex

viveksyngh:

Fixes #60


buildshiprun/handler_test.go
```diff
@@ -123,19 +123,19 @@ func Test_GetImageName(t *testing.T) {
...
 		{
-			"Testcase2",
+			"Repository URL with default port for kubernetes",
 			"127.0.0.1:31115",
 			"registry:31115/username/function-name/",
 			"127.0.0.1:31115/username/function-name/",
 		},
 		{
-			"Testcase3",
+			"Repository URL without port",
```
alexellis: This is better than before, but the input URLs are still quite confusing - as I'm not sure what is being tested here. What is GetImageName producing? The name for a push? Maybe we need to work on the function name too

viveksyngh: It is generating image name for the pull from the push image name and  `repository_url` env variable. Should we change it to `GetPullImageName`?

burtonr:
@viveksyngh This is a very old, and small PR that doesn't appear to be able to be merged any more.

screenshot from 2018-11-14 20-37-43

Perhaps we should close this, and re-open a new one? Maybe have a conversation either in the issue, or in Slack about it too.

Waterdrips:
I'm going to close this as its 18 months old

Issue #41, opened 2018-05-02
Optimisation (status): Use token forwarding and status grouping to avoid extra calls


s8sg:
> This suggestion consider `git-status` as a openfaas function for status update

### Group multiple status update in an array
When required to update multiple status at a time, rather than making a separate call for each status, multiple statuses can be grouped and forwarded at once

#### Possible Solution

**1. Accept multiple commit status in `git-status`**
`git-status` can be changed to accept more than one status at a single execution

**2. Sdk to abstract combine multiple status in a single call**
This can be abstracted by the `sdk` for `Status`
Example:
```
status := BuildStatus()

// add status
status.Add(context1, status1)
// add status
status.Add(context2, status2)

// same status object can store the token from earlier call 
status.Report()
}
```  
In the above example status1 and status2 is forwarded at a single call to `git-status`

**3. Avoid status that will be overridden**
Github map status based on context, which overrides status for same context. Which means for a call sequence like below, the status1 will be overridden by status2 in github
```
status := BuildStatus()

// add status
status.Add(context1, status1)
// add status
status.Add(context1, status2)

// same status object can store the token from earlier call 
status.Report()
}
```
To avoid an extra call the grouped status can be mapped against the context, that will avoid extra call for status that going to be overridden, i.e.
```
CommitStatuses map[string]CommitStatus // context -> CommitStatus
```

### Avoid Authentication by Forwarding the Token

Authentication makes an extra Http call for each status update. 
As the function are stateless, it is (not recommended/ impossible) to store token on functions

#### Possible Solution

**1. Return Token on Success**
`auth-token` can be returned from the `git-status` function on authentication so that it can be reused

**2. Status SDK**
status sdk is meant to abstract the call to `git-status` call. `Status` sdk can itself be used for abstract the call to `git-status`.   
It can be used to abstract the `token` reuse mechanism for multiple `status` update from a single execution of a function 
example: 
```
status := BuildStatus()
Loop {
      // add status
      status.Add(..)
      // same status object can store the token from earlier call 
      status.Report()
}
```  

**3. Forward token to FaaS function**

Token can be forwarded from one function to another, so that it can be reused. 
It can be achieved by setting it either in `http_header` or as a part of data
And can be passed while creating the status 
```
status := BuildStatus(os.Getenv(Http_Token)
```

#### Possible Definition for Status sdk
```
const (
        Success = "success"
        Failure = "failure"
        Pending = "pending"
)

// context constant
const (
[...]

Issue #40, opened 2018-05-02
suggestion (status): Create separate openfaas function for status + sdk for status


s8sg:
As suggested by @alexellis 

##### Separate Openfaas Function For Handing Status Update
- Seperation Of concern
- Avoiding Vendor Code in each Function
- Easier Control to enable/disable function

##### Using SDK for status update
- Avoid duplicate codes
- Streamline status definition and error pattern
- Abstraction handling of token and request

Issue #36, opened 2018-04-30
suggestion (status): Use separate Github status context for each function


s8sg:
`stack.yml` can have multiple `function`s. If any one of the function has status `failure`, the overall status should be `failure`

**This is ensured by github:**

github status allow to specify user defined `context`
```
Status: 
          "state": "success",
          "target_url": "status page",
          "description": "The function has deployed as: user-function_name",
          "context": "function_name"
```

In Github status if one `context` status is `failure` the whole status is `failure`

> Additionally, a combined `state` is returned. The `state` is one of:
> 
> failure if any of the contexts report as `error` or `failure`
> pending if there are no statuses or a context is `pending`
> success if the latest status for all contexts is `success`

Please check [here](https://developer.github.com/v3/repos/statuses/) for more details

**Possible Implementation:**
We can get function specific status in the `github status` by specifying the `context` as `function-name`. 

Once the `yaml` is parsed in `git-tar` the `function` status need to be shown as `pending`
and once `buildshiprun` as successfully deployed the function it need to be shown as `success`
if any error occurs the status should be `failure` 

i.e. for a stack which include `function-1` and `function-2` 
the status will be updated as:
```
// For pending
Status: pending 
function-1: pending
function-2: pending

// For one pending
Status: pending 
function-1: success
function-2: pending

// For success
Status: success
function-1: success
function-2: success

// For failure
Status: failure
function-1: failure
function-2: success
```

**Deal with error before the yaml is parsed:**

Error might happen before the `stack.yml` is even parsed, for example `stack.yml is not present`. We need another separate context before even `stack.yml` is parsed
We can have a context named `Stack Deploy`. 
`Stack Deploy` status can be shown as `pending` as soon as a push event is accepted in `gh-push`
`Stack Deploy` status can be `success` once the yaml is parsed
`Stack Deploy` is failure in case of any `error` in between 

i.e. for a stack which include `function-1` and `function-2` 
the status will be updated as:

##### Stack Deploy (Pending)
<img width="961" alt="screen shot 2018-05-02 at 6 03 37 pm" src="https://user-images.githubusercontent.com/7374058/39561167-e8e38302-4ede-11e8-82e1-ff7add8dd0f2.png">

##### Stack Deploy (Failed)
<img width="954" alt="screen shot 2018-05-03 at 2 38 38 pm" src="https://user-images.githubusercontent.com/7374058/39561295-c094e11a-4edf-11e8-9815-6252865bf4b2.png">

[...]

PR #63, opened 2018-05-28
Add git-status function and status api in sdk (Updated in PR #83)

s8sg:
This commit adds git-status function to handle commit status updates,
along with the below changes
- sdk add status update api to abstract the call to git-status, append
multiple statuses and re-use auth token
- updated sdk is vendor for other functions
- moves status update vendor code from buildshiprun to git-status
- function specific status and overall status for `stack-deploy` by separate contexts 
- pending status for each context

Fix Issue:
#41 #36 #40 

**Overview:** 
- Separate context for each function deploy + context for stack deploy 
- Separate Openfaas Function For Handing Status Update `git-status`
- Using SDK for status update `sdk/statuses.go`
- Using reuse of token (from sdk on a single invocation)
- Using grouping of Status and accept multiple status at `git-status`

**How status gets modified:** 

##### Stack example:
```
provider:
  name: faas
  gateway: http://127.0.0.1:8080

functions:
  regex_go:
    lang: go
    handler: ./regex_go
    image: s8sg/regex_go

  regex_py:
    lang: python
    handler: ./regex_py
    image: s8sg/regex_py
``` 

screen shot 2018-05-02 at 6 03 37 pm

screen shot 2018-05-04 at 12 33 21 pm

screen shot 2018-05-04 at 12 33 28 pm

screen shot 2018-05-04 at 12 33 36 pm

### Links: 
* passed

| context |     url |
| --- | --- |
| stack-deploy | gateway_url: https://7d86c827.ngrok.io/ui/   |
|                        |    pretty url: https://s8sg.7d86c827.ngrok.io     |                    
| regex_go              | gateway url : https://7d86c827.ngrok.io/function/s8sg-regex_go / |
|                        |    pretty url: https://s8sg.7d86c827.ngrok.io/regex_go |
| regex_py              | gateway url : https://7d86c827.ngrok.io/function/s8sg-regex_py / |
|                        |    pretty url: https://s8sg.7d86c827.ngrok.io/regex_py |

* failed/pending

| context |     url |
| --- | --- |
| stack-deploy | git repo   |                    
| regex_go              | git repo |
| regex_py              | git repo |

I have:

- [  ] added unit tests


alexellis:
Hey @s8sg I merged a PR from another contributor and now we'll need a rebase.. sorry 😄

s8sg:
@alexellis I have rebased the same, please check if you can merge the same. Thanks :)

alexellis:
@s8sg please can you confirm what this change is doing and how it's been tested? Positive + negative cases? It seems like a large change for me to merge just before a demo over the next two weeks so I'd like to be sure.

s8sg:
@alexellis This PR works fine 

derek couldn't read the `private-key` from `/var/openfaas/secrets`. as the secrets are mounted at `/run/secrets`
On the code, I'm using  `/var/openfaas/secrets`

s8sg:
I don't see any issues on my deployment
Can you please sent the `git-status` logs ?

alexellis:
@s8sg I've only just realized how big this PR is.

I have given this feedback several times for your PRs.

> Please do only one thing at a time

This changes the whole design of the status reporting as well as adding multiple statuses which makes it very hard for me to review, test and certify. I also have to approve the design changes.

I see the design change is proposed here #41, but it wasn't approved or reviewed so I wouldn't expect it to be part of this PR. The goal of this PR was just to report multiple statuses - one per function in the stack.

Listed changes:

* sdk add status update api to abstract the call to git-status, append
* multiple statuses and re-use auth token
* updated sdk is vendor for other functions
* moves status update vendor code from buildshiprun to git-status
* function specific status and overall status for stack-deploy by separate contexts
* pending status for each context

This is far too much change for a single PR, especially when it affects so much of the code.

I realize the original PR couldn't be merged as quickly as we wanted due to needing to keep the environment stable for demos, but we didn't need to do all this work in one shot. It works far better when these changes are incremental.

Let me take a deeper look at this code change and get back to you.

alexellis:
This is the only work which should be in the PR - https://github.com/openfaas/openfaas-cloud/issues/36

Alex

git-tar/function/handler.go
```diff
@@ -24,29 +24,40 @@ func Handle(req []byte) []byte {
 		os.Exit(-1)
 	}
 
+	statusEvent := sdk.BuildEventFromPushEvent(pushEvent)
+	status := sdk.BuildStatus(statusEvent, "")
```
alexellis: I don't know what `""` signifies from reading the code. Can you explain please?

sdk/events.go
```diff
@@ -14,6 +21,77 @@ type PushEvent struct {
...
+	info.Repository = pushEvent.Repository.Name
+	info.Sha = pushEvent.AfterCommitID
+	info.URL = pushEvent.Repository.CloneURL
+	info.InstallationID = pushEvent.Installation.ID
+
+	return &info
+}
+
+// function to build Event from Environment Variable
+func BuildEventFromEnv() (*Event, error) {
```
alexellis: I don't think this belongs in the status package, this is not generic because the HTTP headers may be different between functions. This should remain in the git-tar function where it is used. It is also only called once which shows it should remain in its own package (the function)

git-tar/function/handler.go
s8sg: The function definition is
```go
func BuildStatus(event *Event, token string) *Status
``` 
It allows to pass and reuse an auth token if available. Currently, we are **not passing** `authtoken` between consecutive functions although the SDK allows to do it.

sdk/events.go
```diff
@@ -14,6 +21,77 @@ type PushEvent struct {
...
+	info.Sha = os.Getenv("Http_Sha")
+	info.URL = os.Getenv("Http_Url")
+	info.InstallationID, err = strconv.Atoi(os.Getenv("Http_Installation_id"))
+	info.Environment = GetEnv(info.Service)
+	info.Secrets = GetSecret(info.Owner, info.Service)
+
+	return &info, err
+}
+
+func GetEnv(service string) map[string]string {
```
alexellis: I couldn't find any usages of this function

sdk/events.go
```diff
@@ -14,6 +21,77 @@ type PushEvent struct {
...
+	if len(envStr) > 0 {
+		envErr := json.Unmarshal([]byte(envStr), &envVars)
+		if envErr != nil {
+			log.Printf("error un-marshaling env-vars for function %s, %s", service, envErr)
+		}
+	}
+	return envVars
+}
+
+func GetSecret(owner, service string) []string {
```
alexellis: This is implementation specific to the function that uses it. Please move it back to its original package.

alexellis:
In addition see that `gh-push` is renamed `github-push` along with `git-event` renaming to `github-event`.

sdk/events.go
s8sg: This is not additionally added for `status` package but it is defined in package `sdk` at `event.go`. And its being used from `buildshiprun`  not `git-tar`

sdk/events.go
s8sg: At line [66](https://github.com/openfaas/openfaas-cloud/pull/63/files/acc57be435ed3804e9aa5ce30d526b4e5e8b57a9#diff-2341a4ae5b399ddbee1e0564106f04a9R66)
```bash
info.Environment = GetEnv(info.Service)
```

viveksyngh:
Sure I will test it but there are still conflicts in this PR. Can you resolve the conflicts and push the changes?

viveksyngh:
I tested it on my local end to end for one deployment. It seems to be working.  But there are so many changes, how do we test so many changes with positive and negative cases? @s8sg 

screen shot 2018-06-28 at 5 57 49 pm

s8sg:
@viveksyngh I have added few more test cases to cover the Status SDK and `github-status` functions. 

You can test the success/Failure manually with [https://github.com/s8sg/regex_go](https://github.com/s8sg/regex_go)

#### Success
Make some nonbreaking change say `README.md` and it should have the commit progress as described on the PR
> The stack deploy, regex_go and regex_py should be successful   
> overall status must be a success 

#### Failure tar creation
change the handler name in `stack.yml`
```yaml
  regex_py:
    lang: python
    handler: ./regex_invalid
    image: s8sg/regex_py
```
> The stack deploy fails 
> overall status will be a failure

Fix the `stack.yml`

#### Failure one function
change the go code `regex_go/handler.go`, to make it fail deliberately    
remove the import 
```go
	"encoding/json"
```
> In this case the stack deploy will be successful 
> regex_py will be successful
> regex_go will be failed
> overall status will be a failure

#### Nothing should be stuck on Pending
None of the statuses should get stuck on pending, they should either fail or pass

I think this is enough to test the functionalities

You can check the `sdk` validation 

#### Check if the public key is used from provided name
Create a new secret
```bash
docker secret create new-private-key <your_private_key_file>.pem
```
and change the secret name at `github.yml`
```
private_key: new-private-key
```
Redeploy and Commit Non-Breaking Change

#### Check cases when status update fails
Create a new secret
```bash
echo "invalid key" > invalid
docker secret rm new-private-key
docker secret create new-private-key invalid
```
Redeploy and Commit Non-Breaking Change
It will create the error in the buildshiprun or in other function as:
```bash
invalid auth token received, token : ( <error string> )
```

> Github-status return the auth token, as it put it in stdout, the `combine_output` must be disabled to reuse the token

#### Reuse authtoken for multiple calls from same function
from `docker service logs github-status` find `reusing provided auth token`
```bash
2018/06/28 15:00:26 Forking fprocess.
2018/06/28 15:00:26 Query  
2018/06/28 15:00:26 Path  /function/github-status
2018/06/28 15:00:27 stderr: 2018/06/28 15:00:26 auth token is created
2018/06/28 15:00:26 Status: success, Context: regex_py, Service: s8sg-regex_go, GitHub AppID: 138863, Repo: regex_go, Owner: s8sg
2018/06/28 15:00:27 Duration: 1.256224 seconds
v1.4fec260fb9db97c97eba6dbadb5da0d12d32a702
2018/06/28 15:00:27 Forking fprocess.
2018/06/28 15:00:27 Query  
2018/06/28 15:00:27 Path  /function/github-status
2018/06/28 15:00:28 stderr: 2018/06/28 15:00:27 reusing provided auth token
2018/06/28 15:00:27 Status: success, Context: stack-deploy, Service: s8sg-regex_go, GitHub AppID: 138863, Repo: regex_go, Owner: s8sg
2018/06/28 15:00:28 Duration: 1.005955 seconds
v1.788f59c2f1a42b64aeab1cbd593afd6e93f56a22
```

alexellis:
I agree with @viveksyngh, but I want to say I appreciate your creativity and ideas for improving the solution. We want your changes, but is there any way we can split them up into separate, smaller chunks that can be integrated, tested and released more incrementally?

I'm also still seeing an error saying: This branch cannot be rebased due to conflicts

s8sg:
@alexellis I checked the extra changes that have been added on the base implementation
   
**Sending Multiple Status at once**
> We are using it in buildshiprun. And it is needed. The current way the feature is implemented it adds the status mapped by the status Context (stack/<func_name>) from `status.AddStatus()`  
> The reason it uses a map by context, so that it overrides the old status with the latest one.
```go
status.CommitStatuses[context] = CommitStatus{Status: state, Description: desc, Context: context
```
We won't have much change in LOC even if we get rid of the feature

**Reusing Auth-Token for consecutive calls from a Function**
> The `ValidToken()` call is used by the `github-status` to check if the returned token from lib is valid 
> As the call already being used otherwise and test cases are written, If we want to get rid of the code to reusing authtoken in current implementation we have to reduce only the below part in `status.Report()`
```go
if !ValidToken(token) {
		log.Printf(`invalid auth token received, token : ( %s ),
make sure combine_output is disabled for git-status`, token)
		status.AuthToken = ""
} else {
		status.AuthToken = token
}
``` 

In both of the above cases, I can have a separate PR but It will be more additional work to make them separate. Personally, I would like to have it in the same PR

   
**Forwarding/ReUsing Auth-Token between Function**
> This will need a change in Event Structure in `sdk`. I agree its a big change and I intentionally haven't added the feature in the PR

PR #67, opened 2018-06-04
[WIP] Load SealedSecrets into cluster

bartsmykla:
Signed-off-by: Bart Smykla <bartek@smykla.com>

## Description

If you deploy your function with secrets.yml in [SealedSecret's](https://github.com/bitnami-labs/sealed-secrets) format it will be deployed into your cluster (k8s) with your function, so you can save secret inside repository without worry that it'll be exposed (it's encrypted using asymmetric keys, which private one is stored inside your cluster)

## How Has This Been Tested?
I have tested it on minikube and my small k8s's cluster on aws (1x c5.xlarge + 2x t2.medium)

## Checklist:

I have:

- [ ] added unit tests


alexellis:
General tips for commit messages -

(don’t exceed 72 chars for any line - for the subject 50-72 chars is the upper limit for GitHub / git log --oneline)

```
Start first line with a verb and a capital letter

Then leave a blank line and for each line don't exceed
72 characters. Explain what you've done and why it
was needed.

And mention how it was tested too.
```

[More info here](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#commit-messages)

alexellis:
It seems odd that `git-tar` is vendoring itself.

```
git-tar/vendor/github.com/openfaas/openfaas-cloud/git-tar/function/ops.go
```

There is a setting for this in `dep` @ericstoekl and @iyovcheva were using it recently

import-secrets/Gopkg.toml
```diff
@@ -0,0 +1,193 @@
...
+
+[[override]]
+  name = "cloud.google.com/go"
+  revision = "3b1ae45394a234c385be014e9a488f2bb6eef821"
+
+[[override]]
+  name = "github.com/Azure/go-autorest"
+  revision = "1ff28809256a84bb6966640ff3d0371af82ccba4"
+
+[[override]]
```
alexellis: This is way more packages than I expected including jwt-go... how come they are needed here but not in our samples?

alexellis:
FYI having reviewed the change with @ericstoekl we found import-secrets is missing from stack.yml.. that would also need to be added for a merge.

import-secrets/Gopkg.toml
bartsmykla: This is done because of **[client-go]**(https://github.com/kubernetes/client-go/blob/master/INSTALL.md#dep-not-supported-yet): 
> As a workaround, which may or may not be worthwhile, you can specify all client-go dependencies manually as override in Gopkg.toml with the versions listed in Godeps.json, and manually update them when you upgrade client-go version.

bartsmykla:
Should be fine now

git-tar/Gopkg.toml
```diff
@@ -24,3 +24,6 @@
 [[constraint]]
   name = "github.com/openfaas/faas-cli"
   version = "0.4.20"
+
+[prune]
+  unused-packages = true
```
johnmccabe: Would be tidier to raise this change in a separate PR so that all the existing vendored deps that get pruned aren't included in this PR.

johnmccabe: @alexellis is there a project wide view on whether we should prune? It can make development using new packages within a given import a bit messier - bit of an edge case perhaps.

alexellis: Can you explain what you mean? pros/cons of both

PR #94, opened 2018-08-10
HMAC validation for github-status + stack-deploy Status failure if any function fails

s8sg:
## Description
This PR add the below changes:
1. Provide `stack-deploy` Status failure if any function fails
2. Add HMAC validation for `github-status`

#### Change 1
It is tested by complete deploytment of `of-cloud`
Below things are tested:
1. If one Function fails, if `stack-deploy` marked as error
2. If all function fails, if `stack-deploy` marked as error
3. If None fails, if `stack-deploy` marked as success

Stack Deploy Failure Example:
screenshot 2018-08-10 at 6 05 06 pm

#### Change 2
It is tested by complete deploytment of `of-cloud`
Below things are tested:
1. If hmac verify is disable
2. If hmac verify is enable
3. If invalid hmac (verify github-status log)
4. If invalid hmac (verify callers failure log )

## Checklist:

I have:
- [X] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [X] added unit tests

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -42,7 +42,6 @@ func Handle(req []byte) string {
 	log.Printf("%d env-vars for %s", len(event.Environment), serviceValue)
 
 	status := sdk.BuildStatus(event, sdk.EmptyAuthToken)
-	status.SetHmacKey(os.Getenv("github_webhook_secret"))
 
 	reader := bytes.NewBuffer(req)
 
@@ -309,7 +308,13 @@ func reportStatus(status *sdk.Status) {
 
 	gatewayURL := os.Getenv("gateway_url")
 
-	_, reportErr := status.Report(gatewayURL)
+	hmacKey, keyErr := sdk.ReadSecret("github-webhook-secret")
+	if keyErr != nil {
+		log.Printf("failed to load hmac key for status, error " + keyErr.Error())
+		return
+	}
+
+	_, reportErr := status.Report(gatewayURL, hmacKey)
 	if reportErr != nil {
 		log.Printf("failed to report status, error: %s", reportErr.Error())
 	}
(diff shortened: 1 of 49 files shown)


alexellis:
Incorrect secret name - it uses dashes and not underscores.

s8sg:
@alexellis It's fixed.

alexellis:
Great, you just need to rebase now as I can see merge conflicts.

alexellis:
I'm not keen on the utility method added named `status.SetHmacKeyFromSecret`

s8sg:
Can you suggest an alternative, otherwise I'll keep it same as it was `status.SetHmacKey()` and handle the secret at function.

s8sg:
@alexellis Done with rebase. I'm doing the `ReadSecret` outside of sdk

buildshiprun/handler.go
```diff
@@ -42,6 +42,7 @@ func Handle(req []byte) string {
 	log.Printf("%d env-vars for %s", len(event.Environment), serviceValue)
 
 	status := sdk.BuildStatus(event, sdk.EmptyAuthToken)
+	status.SetHmacKey(os.Getenv("github_webhook_secret"))
```
alexellis: I would have expected to see the key not set like this, but just passed in the final step where it "sends it" or "submits it".

alexellis: i.e. your helper method named reportStatus

```
reportStatus(status, key)
```

s8sg:
Please consider that either way theres no way to avoid changes in the
status sdk

buildshiprun/handler.go
s8sg: `status.Report(gateway, key)` is the new way. The helper method `reportStatus()` reads the key and forward it.

alexellis:
I think the next step is to refactor out the HTTP-calling/request-constructing code from this function so that we can use it more generically.

PR #95, opened 2018-08-11
Allow for mixed case github usernames

### LINKED ISSUE #54: Proposal:  Case insensitive username handling
We should handle usernames as case insensitive from the CUSTOMERS File

```
Affix
affix
aFfIx
```

should all match to affix

gh-push handler.go line 62

```go
for _, customer := range customers {
	if customer == pushEvent.Repository.Owner.Login {
		found = true
	}
}
```

should be

```go
for _, customer := range customers {
	if strings.EqualFold(customer, pushEvent.Repository.Owner.Login) {
		found = true
	}
}
```

burtonr:
## Description
This change allows customers with mixed case GitHub usernames
to build and deploy functions.

Resolves #54 

Signed-off-by: Burton Rheutan <rheutan7@gmail.com>

## How Has This Been Tested?

Created a new test account @BurtonRTestAccnt with mixed case, installed my OpenFaaS-Cloud app there, created a repo for a function, and pushed to master.

Verified by watching the logs that all steps were successful. Completed the verification by loading the dashboard and seeing the function built. Clicking the link calls the function as expected.

Others are able to test this by either building the new function images themselves, or updating the `stack.yml` for the following functions:
* github-push -> `burtonr/github-push:0.5.0-rc1.4`
* git-tar -> `burtonr/of-git-tar:0.7.0-rc1.1`
* list-functions -> `burtonr/list-functions:0.4.2-1`
> apparently, I'm bad at naming conventions... sorry about that!

## Checklist:

I have:

- [ ] updated the documentation if required **not required for this change**
- [X] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [ ] added unit tests

> Sorry, I did not add unit tests for this, but I could give it a shot in a follow up PR if that's acceptable

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -77,7 +77,7 @@ func Handle(req []byte) string {
 		return auditEvent.Message
 	}
 
-	imageName := result.ImageName
+	imageName := strings.ToLower(result.ImageName)
 
 	repositoryURL := os.Getenv("repository_url")
 	pushRepositoryURL := os.Getenv("push_repository_url")
--- a/garbage-collect/handler.go
+++ b/garbage-collect/handler.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/openfaas/openfaas-cloud/sdk"
@@ -54,7 +55,7 @@ func formatCloudName(name, owner string) string {
 func included(fn *openFaaSFunction, owner string, functionStack []string) bool {
 
 	for _, name := range functionStack {
-		if formatCloudName(name, owner) == fn.Name {
+		if strings.EqualFold(formatCloudName(name, owner), fn.Name) {
 			return true
 		}
 	}
--- a/github-push/handler.go
+++ b/github-push/handler.go
@@ -66,7 +66,7 @@ func Handle(req []byte) string {
 		}
 
 		for _, customer := range customers {
-			if customer == pushEvent.Repository.Owner.Login {
+			if strings.EqualFold(customer, pushEvent.Repository.Owner.Login) {
 				found = true
 			}
 		}
@@ -143,6 +143,11 @@ func getCustomers(customerURL string) ([]string, error) {
 
 		pageBody, _ := ioutil.ReadAll(res.Body)
 		customers = strings.Split(string(pageBody), "\n")
+
+		for i, c := range customers {
+			customers[i] = strings.ToLower(c)
+		}
+
 	}
 
 	return customers, nil
--- a/list-functions/handler.go
+++ b/list-functions/handler.go
@@ -65,7 +65,7 @@ func Handle(req []byte) string {
 
 		for _, fn := range functions {
 			for k, v := range fn.Labels {
-				if k == "Git-Owner" && v == user {
+				if k == "Git-Owner" && strings.EqualFold(v, user) {
 					// Hide internal-repo details
 					fn.Image = fn.Image[strings.Index(fn.Image, "/")+1:]
 					filtered = append(filtered, fn)


alexellis:
Quick thought.. Does this cover of-router?

burtonr:
I'm not sure. I couldn't find reference to of-router in any of the docs. 

It looks like it's only used/deployed when using the Kubernetes deployment. I had only setup and run on Swarm. That is something that would need verified though!

I also couldn't find where, or how of-router is called/used. 

Looking through the of-router `main.go` it looks like it just forwards the request to the `<gateway>/function/<username>-<function name>`. If that's the case, it should work fine as now the username is converted to lower case before any of the build or deploy steps happen. It _should_ be fine...

alexellis:
Rater than mutating the data, I think we need to store the original username in its proper casing and at all times convert to lower for comparisons, formatting etc.

alexellis:
of-router also works with swarm, the current live environment is setup with that. You can just use a host files entry to test it.

Imagine if our domain was `*.case-test.team-serverless.xyz`, then each user would have `username.case-test.team-serverless.xyz`. You'd have `burtonr.case-test.team-serverless.xyz`

So if you don't want to setup Route53/Google DNS or similar you can edit your hosts file

```
this.is.the.ip burtonr.case-test.team-serverless.xyz
```

git-tar/function/handler.go
```diff
@@ -26,6 +27,9 @@ func Handle(req []byte) []byte {
 		os.Exit(-1)
 	}
 
+	// force github username to lowercase to prevent issues downstream
+	pushEvent.Repository.Owner.Login = strings.ToLower(pushEvent.Repository.Owner.Login)
```
alexellis: Not sure we should mutate the value from GitHub, this also invalidates the SHA signature.

burtonr:
Apologies for the delay! I've been a bit busy over the weekend. 

I understand the concern with modifying the data. I went that route to fix it at the earliest possible place so that all the downstream services (and future services) wouldn't need to worry about case checking.

Give me a couple of days, and I'll update this to take the value as-is, and change it, or invariant-ly check it, through all the different services. I'll also need to re-base my test cluster the recent updates that happened since I started this. I'm hopeful for a Wednesday update.

burtonr:
Alright, I got it this time. This changeset looks a lot better than the first attempt! Thanks @alexellis  for pushing for doing it the right way.

I've updated the services to put the values to lower case where it's necessary, rather than forcing it at the source. This way, the repo owner and repo names are passed around and stored as-is, then converted to lower case only when they are being used for image names and validation. I'm much happier with this.

buildshiprun/handler.go
```diff
@@ -38,7 +38,7 @@ func Handle(req []byte) string {
 		Source: "buildshiprun",
 	}
 
-	serviceValue := fmt.Sprintf("%s-%s", event.Owner, event.Service)
+	serviceValue := strings.ToLower(fmt.Sprintf("%s-%s", event.Owner, event.Service))
```
alexellis: We don't need to change the service name, this is unrelated to GitHub usernames.

alexellis:
Hey a rebase is needed now @burtonr 

I'd also like to know that the dashboard still works and of-router

Thanks,

Alex

burtonr:
@alexellis I've done the rebase. This is ready to go now.

I've verified that the dashboard does still function as expected. Using either lower case, or mixed case in the query string brings up the same page, and the "Endpoint" `Link` routes correctly to the function.

For the of-router, I wasn't able to get that functional in my setup, but following the readme, and watching the logs, it was attempting to route to the correct path. I'm confident that of-router will work as expected.

alexellis:
I can still see the username getting mutated in one place. https://github.com/openfaas/openfaas-cloud/pull/95/files#r209429701

burtonr:
@alexellis That must have gotten mixed up in my rebase from earlier. That was not in the function that I had deployed and tested with.

I've updated the commit to remove that data manipulation

PR #97, opened 2018-08-15
Capture logs from of-builder

alexellis:
## Description

A struct is now returned as JSON from of-builder containing an
array of lines from buildkit and the imagename/status.

Tested locally with Docker for Mac - I received logs on a
positive/negative build.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -101,9 +101,13 @@ func Handle(req []byte) string {
 		msg := "Unable to build image, check builder logs"
 		status.AddStatus(sdk.StatusFailure, msg, sdk.BuildFunctionContext(event.Service))
 		reportStatus(status)
-		log.Fatal(msg)
+
 		auditEvent.Message = fmt.Sprintf("buildshiprun failure: %s", msg)
 		sdk.PostAudit(auditEvent)
+
+		log.Printf("of-builder result: %s, logs: %s\n", result.Status, strings.Join(result.Log, "\n"))
+
+		log.Fatal(msg)
 		return msg
 	}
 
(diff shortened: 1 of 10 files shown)


sdk/build.go
```diff
@@ -0,0 +1,5 @@
+package sdk
+
+type Build struct {
```
kvuchkov: Isn't BuildResult a more suitable struct for the sdk, as it is returned by the build service?

alexellis: Yes for sure

of-builder/Dockerfile
```diff
@@ -8,18 +8,17 @@ RUN go build -o /usr/bin/of-builder .
 
 FROM alpine:3.7
 
-RUN addgroup -S app && adduser -S -g app app
-RUN mkdir -p /home/app
+RUN addgroup -S app && adduser app -S -G app \
```
kvuchkov: Adding the user to a group is causing an "operation not permitted" when untar tries to lchown in the tmp folder, e.g. "/tmp/buildctx584615953/context".

of-builder/main.go
```diff
@@ -22,12 +22,23 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var (
+	lchownEnabled bool
```
kvuchkov: If we revert the group of the user, should we remove the flag?

buildshiprun/handler.go
```diff
@@ -365,3 +378,11 @@ func getRegistryAuthSecret() string {
 	}
 	return ""
 }
+
+// BuildResult represents a successful Docker build and
+// push operation to a remote registry
+type BuildResult struct {
```
kvuchkov: Do you think we need to extract this struct in sdk?

alexellis: Sure

of-builder/main.go
alexellis: We may need it again for testing, I'll leave for now but unused.

alexellis:
Ugh

alexellis:
Messed this PR branch up and merged the same via #102

PR #106, opened 2018-08-20
Adds docker login tips for Mac and Windows to docs/README.md

kvuchkov:
docker login uses credentials store by default on Mac and Windows. It will be very convenient if we include a simple instruction how to work around this and be able to deploy openfaas-cloud localy with ease.

Signed-off-by: Kiril Vuchkov <kirilvuchkov@gmail.com>

## How Has This Been Tested?
I've tried the instructions myself on Mac. I haven't tested this on Windows, however it should be the same.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -175,6 +175,7 @@ This populates `~/.docker/config.json` which is used in the builder:
 }
 
 ```
+> Note: the auth string above is using an example value and is not a real authentication string.
 
 #### For Kubernetes
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -125,15 +125,56 @@ You will need to edit `stack.yml` and make sure `buildshiprun_limits_swarm.yml` 
 
 ### Deploy your container builder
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -125,15 +125,56 @@ You will need to edit `stack.yml` and make sure `buildshiprun_limits_swarm.yml`
...
+	"auths": {
+		"https://index.docker.io/v1/": {
+			"auth": "asdf12djs37ASfs732sFa3fdsw=="
+		}
+	},
+	"HttpHeaders": {
+		"User-Agent": "Docker-Client/18.06.0-ce (darwin)"
+	}
+}
+
```
alexellis: > Note: the auth string above is using an example value and is not a real authentication string.

I'd add the above, because people have (and will) complain that we've shared our login password etc

kvuchkov: OK, I haven't though about anyone would think we share actual credentials. Good point - I am adding it.

alexellis:
LGTM, thank you Kiril.

Issue #7, opened 2018-04-04
Support for private repos


tpiskula:
First of all, thx for your amazing work! This is exactly what I've been looking for.
I have this dream of a netlify-like workflow where you just push function code a repo and everything else just works :D 
What's the roadmap for openfaas-cloud?
It would be awesome If it could support private github repos, eg automatically add a deploy key to the repo and use that in the clone step, are there any plans in that direction?
I would be glad to help implementing!

alexellis:
Hi @tpiskula the roadmap is represented in GitHub issues at the moment - there's lots of ways you can get involved with testing and building out the project. I think the idea of private repos would be a great addition to OpenFaaS Cloud.

Here's instructions for how to join Slack where we also have a private demo of a shared/hosted OpenFaaS cloud - http://docs.openfaas.com/community/

alexellis:
I've looked into private repos and I think there is a good option for pulling / cloning. If you add a "Deploy Key" to the repo then you can clone over SSH. This does need a public/private key-pair for every repo supported - unfortunately GitHub won't allow one deploy key to be added to multiple repos.

I think this means that OpenFaaS Cloud needs to generate a key/pair for each user and repo - this could potentially be done via the API if OpenFaaS Cloud is given repository administration access via https://github.com/google/go-github/blob/master/github/repos_keys.go

We'd also need to check whether the key already existed and use a name to index it.

alexellis:
[OpenFaaS private repo flow.pdf.pdf](https://github.com/openfaas/openfaas-cloud/files/2145456/OpenFaaS.private.repo.flow.pdf.pdf)

Attaching workflow/design.

We will need to generate private/public key-pairs for each repo we manage. These have to be stored for future use and the public key must be installed on the GitHub repo needing additional OAuth Scopes.

tpiskula:
Workflow looks good!
I imagine it something like:
Authorize Github Account -> Select Repo -> If repo is private, install Deploy Key via API and save it somewhere -> Run CI/CD, run CI/CD on every push (maybe allow filter on branch?)
Where would you suggest to store the generated keys?
In a k8s secret?

alexellis:
This part (`Authorize Github Account`) is done via the installation event when you click to install the GitHub app on your repo in the GitHub UI.

Run CI/CD is triggered on the push.

I think we have to store them in K8s secrets, possibly SealedSecrets could be useful for us too.

The other alteration needed is to:

- connect installation event via github-event function  
[...]

PR #120, opened 2018-08-31
Added option to clone private respositories

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/issues/7

Signed-off-by: Bart Smykla (VMware) <bartek@smykla.com>

## Description
We have github's auth token for displaying statuses so we could
use the same approach and generate auth token just for cloning
private repositories.

I moved GetPrivateKeyPath function from github-status, and
placed it inside sdk/auth because we can reuse itside git-tar
to clone private repository. I have created tests for
that function also.

I like to create pure functions so I created an abstraction
of authToken, which has method GetToken inside. In the future
if we will have auth token already, we could reuse it,
but probably we'll have to add mechanism to check if token
maybe should be generated again (it's live ended).
Because of that abstraction, I could create a tokener interface
which is used inside getRepositoryURL function. It helps with
creation of unit tests.

getRepositoryURL function will return the same clone URL
for repository if it's public, but if it's private,
it'll get a token first, then it'll put installationID:authToken
to the url so we can clone private repository
(ie. https://installationID:authToken@github.com/organization/repository)

I added .idea directory to the .gitignore because I'm using
jetbrains' IDEs also.

## How Has This Been Tested?

- Changed repository to private and pushed changes to it
- Created new repository (private from beginning) and connected it to OpenFaaS Cloud
- Changed repository from Private to Public and pushed changes to it
- Created new public repository and connected it to OpenFaaS Cloud 

EKS cluster 2x t2.medium

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/git-tar/Gopkg.toml
+++ b/git-tar/Gopkg.toml
@@ -33,10 +33,14 @@
   name = "github.com/openfaas/openfaas-cloud"
   version = "0.5.2"
 
-[prune]
-  go-tests = true
-  unused-packages = true
-
 [[constraint]]
   name = "github.com/alexellis/hmac"
   version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/alexellis/derek"
+  version = "0.6.1"
+
+[prune]
+  go-tests = true
+  unused-packages = true
--- a/git-tar/function/ops.go
+++ b/git-tar/function/ops.go
@@ -10,15 +10,18 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/openfaas/faas-cli/schema"
 
-	hmac "github.com/alexellis/hmac"
+	"github.com/alexellis/derek/auth"
+	"github.com/alexellis/hmac"
 	"github.com/openfaas/faas-cli/stack"
 	"github.com/openfaas/openfaas-cloud/sdk"
 )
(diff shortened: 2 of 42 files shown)


ivanayov:
@bartsmykla You can add this config to `Gopkg.toml`: 
```
[prune]
  go-tests = true
  unused-packages = true
```

This will remove the non-sdk vendor files.

bartsmykla:
@ivanayov it is there

git-tar/function/ops.go
```diff
@@ -165,6 +168,64 @@ func formatImageShaTag(registry string, function *stack.Function, sha string, ow
...
 }
 
+type authToken struct {
+	appID          string
+	installationID int
+	privateKeyPath string
+	token          string
+}
+
+func (t *authToken) GetToken() (string, error) {
```
alexellis: Why is this exposed outside the package?

alexellis: Or rather.. why the capital letter indicating public?

git-tar/function/ops_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if got != expected {
+		t.Errorf("Expected: %s, Got: %s", expected, got)
+	}
+}
+
+func Test_getRepositoryURL_whenRepositoryIsPrivate(t *testing.T) {
+	pe := genPushEvent(12345, "https://foo.bar/baz", true)
+	at := &authTokenMock{}
```
alexellis: Since this is not a mock but a stub/fake let's change the name. test or fakeAuthToken may work for instance.

git-tar/function/ops.go
```diff
@@ -183,7 +244,19 @@ func clone(pushEvent sdk.PushEvent) (string, error) {
 		return "", fmt.Errorf("cannot create user-dir: %s", userDir)
 	}
 
-	git := exec.Command("git", "clone", pushEvent.Repository.CloneURL)
+	at := &authToken{
```
alexellis: Can we call `at` something like `authToken` and `authToken` something else if needed. `githubAuthToken` / `installationAuthToken`?

git-tar/function/ops.go
```diff
@@ -165,6 +168,64 @@ func formatImageShaTag(registry string, function *stack.Function, sha string, ow
...
+	t.token = token
+
+	return token, nil
+}
+
+func (t *authToken) GetInstallationID() int {
+	return t.installationID
+}
+
+type tokener interface {
```
alexellis: Not sure about the name `tokener`

git-tar/Gopkg.toml
```diff
@@ -40,3 +40,7 @@
 [[constraint]]
   name = "github.com/alexellis/hmac"
   version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/alexellis/derek"
+  version = "0.5.5"
```
alexellis: The latest release here will mean we can remove much of the new dependencies 👍

git-tar/function/ops.go
bartsmykla: What name do you suggest?

alexellis: What's it's job?

PR #127, opened 2018-09-04
Add current timestamp if no vertex timestamp

s8sg:
This commit adds current time as a format
```
2018-09-03 16:27:09.580093614 +0000 UTC
```
In case the vertex timestamp v.Started is not available

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/of-builder/main.go
+++ b/of-builder/main.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/docker/docker/pkg/archive"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


of-builder/main.go
```diff
@@ -168,9 +169,12 @@ func build(w http.ResponseWriter, r *http.Request) ([]byte, error) {
 				if v.Completed != nil {
 					msg = fmt.Sprintf("v: %s %s %.2fs", v.Started, v.Name, v.Completed.Sub(*v.Started).Seconds())
 				} else {
-					startedVal := "(no timestamp available)"
+					startedVal := ""
 					if v.Started != nil {
 						startedVal = v.Started.String()
+					} else {
+						tnow := time.Now()
+						startedVal = tnow.Format("0000-00-00 23:59:59.999999999 +0000 UTC")
```
alexellis: Not sure we need the format piece. This is just using `.String()` on a `time.Time` struct. Can you update please?

alexellis:
I've given some feedback.

Please can you also add a method that formats the timestamp to less precision? We should use a shorter value for the seconds and I think we can drop the timezone.

of-builder/main.go
s8sg: `tnow.Format` formats the time in less precision.

alexellis:
I'm thinking we should use `time.RFC3339` for brevity.

of-builder/main.go
s8sg: @alexellis sorry for picking this one late. Do you mean to apply `time.RFC3339` to format time in all logs `v.Started` and the `tnow` ?

alexellis:
Hi @s8sg if you provide some test output you will see that the code added for `.Format(time.RFC3339)` will only be applied to one of the dates and now they will be mis-matched.

Can you update this PR so that all dates outputted into the logs are done in the same format? i.e. `.Format(time.RFC3339)`

This should be a 5-10 min re-work plus a quick manual verification with the builder to show it's working as expected.

Alex

alexellis:
Derek close: finishing work in #207, @s8sg messaged me to say he can't finish the open PRs right now.

PR #134, opened 2018-09-05
Add makeHandler() test and moving host in own variable

martindekov:
Adding test for makeHandler function which covers everything
but the closing of the body and moving the host out of
the direct path since if `Host:` is unset it will cause panic
due to trying to access non existing memory

## Description

Added test for makeHandler() function which covers everything but the closing of the response body
and moved out handling of `r.Host` out of the path creation since in cases where `Host:` was unset it tried to access non existing address which resulted in panic which prevented the test from working properly

## How Has This Been Tested?

Added the test. N.A.

## Checklist:

I have:

- [x] added unit tests


router/main.go
```diff
@@ -40,17 +40,19 @@ func main() {
 
 func makeHandler(c *http.Client, timeout time.Duration, upstreamURL string) func(w http.ResponseWriter, r *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
-
+		var host string
 		if len(r.Host) == 0 {
```
alexellis: Did this guard not work? Was `r.Host` actually `nil`?

router/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+			t.Fatalf("\nRequest error: `%v`\n", err)
+		}
+		rec := httptest.NewRecorder()
+		c := &http.Client{
+			Timeout: test.Timeout,
+		}
+		handlerFunc := makeHandler(c, test.Timeout, test.UpstreamURL)
+		handlerFunc(rec, req)
+		res := rec.Result()
+		defer res.Body.Close()
```
alexellis: The recorder might be able to provide this without reading the body bytes in the more verbose way.

router/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+		Host               string
+		RequestURI         string
+		ExpectedSuffix     string
+		UpstreamURL        string
+		Timeout            time.Duration
+		ExpectedStatusCode int
+	}{
+		{
+			Title:              "Right configuration with status 500 unrelated to function",
+			Host:               "martindekov.example.xyz/",
```
alexellis: `RequestHost`? or maybe `requestURI`?

router/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+			RequestURI:         "myfunction",
+			ExpectedSuffix:     "martindekov-myfunction",
```
alexellis: What about the whole expected upstream URL?

router/main.go
martindekov: `r.Host[0:strings.Index(r.Host, ".")]`  panics when r.Host == "" maybe even the thing I provided is not guard enough for example what happens when we have string that does not have `.` I haven't really researched that case

router/main_test.go
martindekov: Well the response body is not as expected 
```
Proxying to: http://localhost:8080/function/martindekov-myfunction
... ```
we only print.. that but rather
```
unable to scale function [martindekov-myfunction], err: error scaling HTTP code 500, http://faas-swarm:8080/system/scale-function/martindekov-myfunction
```
because we resent the request but I can find a way to send `Proxying to: ...` message in the body of the response

**Comment by martindekov** [comment]
Referenced the comments about host naming, missed that recorder has `Code` and `Body` so referenced that also now I don't use function to get the content etc.

File: router/main.go (hunk 1 above)
Comment (alexellis) [reply router/main.go]: Sounds good. Do your tests cover both scenarios?

File: router/main_test.go [hunk 5]
```diff
@@ -0,0 +1,90 @@
...
+			UpstreamURL:        "",
+			Timeout:            time.Second * 30,
+			ExpectedStatusCode: http.StatusServiceUnavailable,
+		},
+		{
+			Title:              "Everything configured but RequestHost resulting in status 400",
+			RequestHost:        "",
+			RequestURI:         "/myfunction",
+			ExpectedSuffix:     "-myfunction",
```
alexellis: I just wonder if it's simpler for us to have this like:

```
Title
inputURL
expectedUpstreamURL
```

And to leave it at that?

router/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+			Timeout: test.Timeout,
+		}
+		handlerFunc := makeHandler(c, test.Timeout, test.UpstreamURL)
+		handlerFunc(rec, req)
+		if rec.Code != test.ExpectedStatusCode {
+			t.Logf("\nTest: `%v`\n", test.Title)
+			t.Errorf("Expected status code : `%v` got: `%v`", test.ExpectedStatusCode, rec.Code)
+		}
+
+		read, err := ioutil.ReadAll(rec.Body)
```
alexellis: What about `rec.Result().Body`? It shouldn't require `ioutil`.

router/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+			ExpectedStatusCode: http.StatusBadRequest,
+		},
+		{
+			RequestHost:        "",
+			UpstreamURL:        "http://localhost:8080/",
+			ExpectedStatusCode: http.StatusBadRequest,
+		},
+	}
+
+	timeout := time.Second * 30
```
alexellis: We shouldn't have such a big timeout in a unit test. Set it to 1 s?

router/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for _, test := range config {
+		req, err := http.NewRequest(http.MethodPost, "/", nil)
+		if err != nil {
+			t.Fatalf("Request error: `%v`\n", err)
+		}
+
+		URI := strings.TrimPrefix(test.UpstreamURL, "http://localhost:8080")
```
alexellis: hard-coded localhost should be moved out up higher please.

router/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+		req.RequestURI = URI
+		req.Host = test.RequestHost
+
+		rec := httptest.NewRecorder()
+		c := &http.Client{
+			Timeout: timeout,
+		}
+
+		URL := strings.TrimSuffix(test.UpstreamURL, "myfunction")
```
alexellis: I'd like to see the function name moved up higher or put in the test scenario

router/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+
+		read, err := ioutil.ReadAll(rec.Body)
+		if err != nil {
+			t.Errorf("Unexpected response body read error: `%v`", err)
+		}
+
+		bodyValue := string(read)
+		var ExpectedSuffix string
+		userFromURI := strings.TrimPrefix(URI, "/")
```
alexellis: Seems like we're doing a lot of trimming and checking in the test. This makes it look like we're testing the wrong thing(s) or the wrong level of granularity. Any thoughts?

Normally a test is:

Arrange - some data

Act - call the method passing in the data

Assert - check that the actual vs expected match.

router/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		URL := strings.TrimSuffix(test.UpstreamURL, "myfunction")
+		handlerFunc := makeHandler(c, timeout, URL)
+		handlerFunc(rec, req)
+
+		if rec.Code != test.ExpectedStatusCode {
+			t.Errorf("Expected status code : `%v` got: `%v`", test.ExpectedStatusCode, rec.Code)
+		}
+
+		read, err := ioutil.ReadAll(rec.Body)
```
alexellis: Do you need to call `ioutil.ReadAll` anymore?

router/main_test.go
martindekov: You requested to reduce the number of fields in struct to max 3 that's why I did so much trimming etc. Should I increment the test cases again?

router/main_test.go
alexellis: `rec.Body` is `Body *bytes.Buffer` - correct me if I'm wrong, but I don't think you need to call `ReadAll` on it.

martindekov: That way yeah the response body is `bytes.Buffer` but I need it as `[]byte` to convert it to string later on and compare values. The same approach is in `requesthandler_test.go` in `faas/watchdog`

PR #137, opened 2018-09-06
Add node8-express-template to git-tar function

kenfdev:
Signed-off-by: Ken Fukuyama <kenfdev@gmail.com>

## Description
This explicitely adds a template fetching phase in `git-tar`.  It fetches the default template plus the `node8-express-template`. Since it doesn't make much sense to manage the repos externally, it is hard coded inside the `fetchTemplates` function.

## How Has This Been Tested?

Since i do not have a OpenFaaS Cloud cluster locally to test with, I have only been able to invoke the `fetchTemplates` function directly from a local unit test and check that the default *and* node8-express-template has been successfully pulled.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/git-tar/function/ops.go
+++ b/git-tar/function/ops.go
@@ -42,6 +42,25 @@ func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error
 	return parsed, err
 }
 
+func fetchTemplates(filePath string) error {
+	templateRepos := []string{"https://github.com/openfaas/templates", "https://github.com/openfaas-incubator/node8-express-template.git"}
+
+	for _, repo := range templateRepos {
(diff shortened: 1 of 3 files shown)


git-tar/function/ops.go
```diff
@@ -42,6 +42,25 @@ func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error
 	return parsed, err
 }
 
+func fetchTemplates(filePath string) error {
+	templateRepos := []string{"https://github.com/openfaas/templates", "https://github.com/alexellis/node8-express-template"}
```
alexellis: Are you sure this is on my account still and not the incubator?

kenfdev: Thank you for the review. The value was old, sorry. I've fixed it to the incubator one.

PR #151, opened 2018-09-09
Fix function shortname

### LINKED ISSUE #150: Dashboard: incorrect links for usernames containing (-)
## Expected Behaviour

For the following username we should present the correctly formatted endpoint link for his "hello-go" function.

https://system.o6s.io/overview?user=zhenfeng-zhu

i.e. https://zhenfeng-zhu.o6s.io/hello-go

## Current Behaviour

Due to some code in the dashboard UI calculating the "shortName" variable we are appending part of the username to the endpoint URL causing it to fail to work:

Instead of https://zhenfeng-zhu.o6s.io/hello-go we see: https://zhenfeng-zhu.o6s.io/zhu-hello-go

## Possible Solution

This should be a simple fix to the overview function's JavaScript and needs a dual-fix in the new PR via @kenfdev #148  

1. Use links above or register a new GitHub username containing a dash.
2. Follow endpoint link.

## Context

This will affect people who have one or more dashes in their GitHub username.

Only the UI is affected from what I can tell and not the underlying router, deployment or meta-data.

kenfdev:
Signed-off-by: Ken Fukuyama <kenfdev@gmail.com>

## Description

This fixes the shortname of the function even though it includes a `-`.
It assumes the function `name` is always prefixed with the `user` and replaces that string.

Closes #150 

This fix also needs to be applied to #148 

## How Has This Been Tested?

Since I don't have my own OpenFaaS Cloud cluster, I've only tested this as plain JavaScript (Unit test level)

```js
var user = "zhenfeng-zhu"
var name = "zhenfeng-zhu-hello-go";
var userPrefixRegex = new RegExp("^" + user + "-");
name.replace(userPrefixRegex, ""); // "hello-go"
```

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/dashboard/overview/index.html
+++ b/dashboard/overview/index.html
@@ -119,6 +119,7 @@
                     return 0
                 });
 
+                var userPrefixRegex = new RegExp("^" + user + "-");
                 for(var i = 0; i < data.length; i++) {
                     var index = i;
                     var item = data[index];
@@ -126,10 +127,7 @@
                     var since = new Date(parseInt(item.labels["Git-DeployTime"])*1000);
                     var sinceDuration = moment(since).fromNow();
 
-                    var shortName = item.name;
-                    if(shortName.indexOf("-") > -1) {
-                        shortName = shortName.substr(shortName.indexOf("-") + 1);
-                    }
+                    var shortName = item.name.replace(userPrefixRegex, "");
 
                     var url;
                    
--- a/dashboard/stack.yml
+++ b/dashboard/stack.yml
@@ -9,7 +9,7 @@ functions:
   system-overview:
     lang: go
     handler: ./overview
-    image: alexellis2/overview:0.4.3
+    image: functions/overview:0.4.4
     environment:
       content_type: "text/html"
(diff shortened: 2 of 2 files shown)


kenfdev:
I've implemented the same version for the react dashboard prototype which you can check here:

https://kenfdev.o6s.io/dashboard/zhenfeng-zhu

You can see that the endpoint is fixed.

kenfdev:
Thank you @alexellis . I've bumped both overview and list-functions version. Was that okay?

alexellis:
Did list-functions change?

kenfdev:
> Did list-functions change?
No. Somehow I thought these had same version numbers but I'll revert the list-functions.

> Let's also change the Hub prefix to functions/ to match the root stack file.
Got it

kenfdev:
Done!

alexellis:
Thank you. I'll deploy this now.

alexellis:
LGTM @kenfdev -> https://system.o6s.io/overview?user=zhenfeng-zhu

PR #164, opened 2018-09-12
Auto-detect buildshiprun limits

martindekov:
Removing the two limit files and leaving only one
in which we define the limit with single integer value
then check if "KUBERNETES_SERVICE_PORT" is present
and if so we set suffix of `Mi` by defualt the suffix
is `m`

## Description

Now the `buildshiprun` limit formatting is handled by the function itself if `KUBERNETES_SERVICE_PORT` environmental var is present which suggest we are running on kubernetes rather than swarm. And now we configure the limits in single `buildshiprun_limits.yml` rather than having two for the purpouse.

## How Has This Been Tested?

Tested only on swarm the following two scenarios
```
# cat buildshiprun_limits.yml
environment:
  default_memory_limit: 30
```
and put log into `buildshiprun` to tell me what the limit is set to the logs give
```
buildshiprun.1.vp0mfptfvy1k@FaasSwarm    | 2018/09/12 11:33:32 Memory limit: `30m`
```
now I comment out the file from configuration.
```
...
    environment_file:
     # - buildshiprun_limits.yml
...
```
the logs give
```
...
buildshiprun.1.ylkzogu0d5m9@FaasSwarm    | 2018/09/12 11:44:35 Memory limit: `20m`
...
```
Because the default is set to `20` if unset
## Checklist:

I have:

- [x] added unit tests


buildshiprun/handler.go
```diff
@@ -453,3 +451,17 @@ func getRegistryAuthSecret() string {
 	}
 	return ""
 }
+
+func getMemoryLimit() string {
+	suffix := "m"
+	memoryLimit := os.Getenv("default_memory_limit")
+	if _, present := os.LookupEnv("KUBERNETES_SERVICE_PORT"); present == true {
```
alexellis: I'd like to see this split out into a separate method which is also tested independently. In the future we may want to change the way we find out whether we're in Kubernetes and so having this abstracted would be ideal.

No need to move to the sdk yet though.

martindekov:
It can be gigabyte or something other than megabyte so yeah I will reference that

martindekov:
My main suggestion is sticking to swarm approach with suffix aka. `30m` `30g` etc. I find this more convenient since we don't have CPU limits for now or at least I am not aware of them and changing files in stack.yml is additional configuration step which is not necessary

Other approach is we might create our own convention like `Mg` `Gb` `Kb` `By` etc.

The last approach is stay as it is.

These are my suggestions, so how should I proceed?

alexellis:
My thought was perhaps you can write:

```
function_limit_mb: 128
```

alexellis:
We never need less than 1mb and if we want to supply a "gb" value we can just multiply by 1024.

PR #157, opened 2018-09-10
Force scaling limits. Make them configurable

ivanayov:
With this change buildshiprun command applies scaling limits to
the deployment. The default values are set to 1 for min and
4 for max. They are configurable in the `gateway_config.yml`

Signed-off-by: Ivana Yovcheva <iyovcheva@vmware.com>

Resolves #154

## How Has This Been Tested?

After deployment the labels are available in 
```
{  
   "name":"ivanayov-hello-wrold",
   "image":"ivanayov/ivanayov-faas-cloud-test-hello-wrold:latest-dffc797",
   "invocationCount":0,
   "replicas":1,
   "envProcess":"",
   "availableReplicas":1,
   "labels":{  
      "Git-Cloud":"1",
      "Git-DeployTime":"1536599218",
      "Git-Owner":"ivanayov",
      "Git-Repo":"faas-cloud-test",
      "Git-SHA":"dffc79772e100bee742b4584752165c2885c9e86",
      "app":"ivanayov-hello-wrold",
      "com.openfaas.function":"ivanayov-hello-wrold",
      "com.openfaas.scale.max":"4",
      "com.openfaas.scale.min":"1",
      "com.openfaas.uid":"859184577",
      "faas_function":"ivanayov-hello-wrold",
      "function":"true"
   },
   "annotations":null
}
```

## Checklist:

I have:

- [ ] updated the documentation if required
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [ ] added unit tests

PR #165, opened 2018-09-12
Fix buildshiprun scaling limits defaults

ivanayov:
This fixes buildshiprun handler to set properly default values
for min and max scailing limits. Adds test coverage of how configuration is loaded

Follows up pull #157

### DIFF
--- a/buildshiprun/handler_test.go
+++ b/buildshiprun/handler_test.go
@@ -175,7 +175,6 @@ func Test_getReadOnlyRootFS_override(t *testing.T) {
 		t.Fail()
 	}
 }
-
 func Test_getMemoryLimit_Swarm(t *testing.T) {
 	tests := []struct {
 		title         string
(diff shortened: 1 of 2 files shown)


alexellis:
Can you give me some more context please?

ivanayov:
I committed a bug in #157, by not checking properly the configuration values, which causes the defaults for scaling limit min and max not to be set.

alexellis:
Since this was a bug, can we write a test and additional method to load this config?

ivanayov:
@alexellis  ok, I'll add it

ivanayov:
It's now updated with test

ivanayov:
@alexellis Thanks for the feedback. Your comments are now addressed.

buildshiprun/handler_test.go
```diff
@@ -175,3 +175,29 @@ func Test_getReadOnlyRootFS_override(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func Test_getConfig(t *testing.T) {
+
+	os.Setenv("scaling_max_limit", "")
```
alexellis: Please use separate tests for each case or a test scenario slice like we do in Derek with a name for each case.

ivanayov:
@alexellis Addressed your last comment

bartsmykla:
@ivanayov you have 👍 from me. LGTM

alexellis:
Since Bart gave LGTM, I'll merge now. Thanks for working on the feedback.

PR #170, opened 2018-09-13
Fixes #169

### LINKED ISSUE #169: Build dashboard static assets using a Node.js container
## Expected Behaviour

`make` within the dashboard sub-folder should generate static assets using a container and a bind-mount so that we have a consistent experience for all developers.

## Current Behaviour

Depends on local tooling such as `yarn` which may be at different versions.

## Possible Solution

Use the official Node.js / Alpine container to carry out the build by mounting the source/destination folders. This may need a Dockerfile or may be possible with a simple `docker run`.

1. `cd dashboard`
2. `make`

This should work without `yarn` or `node` on the local system.

jkrol2:
Use node:8-alpine container

Signed-off-by: Jakub Krol <jakub.krol@poczta.fm>

## Description
I switched the command in make to `docker run` command executing building inside official Nodejs container.

## How Has This Been Tested?
I have run `make` command inside dashboard directory

## Checklist:

I have:
- [ x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x ] signed-off my commits with `git commit -s`

### DIFF
--- a/dashboard/Makefile
+++ b/dashboard/Makefile
@@ -1,4 +1,4 @@
 .PHONY: build
 
 build:
-	(cd client; yarn; yarn run build)
+	( docker run -v $(shell pwd):/dashboard node:8-alpine /bin/sh -c "cd dashboard/client && yarn && yarn run build")


alexellis:
Merged. Thanks both.

PR #177, opened 2018-09-14
[WIP] Expose additional metrics

### LINKED ISSUE #162: Function: Expose additional metrics
## Expected Behaviour

We should be able to bubble up r/s for 2xx and non 2xx (error) codes over the last hour for use in the dashboard's detail page.

Without changing the gateway API we can query Prometheus directly for a single named function and bring back data.

Example query that returns the data with the codes (replace figlet in the string):
```
sum(rate(gateway_function_invocation_total{function_name="figlet"}[60m])) by (code)
```

The easiest way to do this may be to create a small function in Go which can query the data for a given function name and return the data in JSON for usage in the new dashboard in #148 

Request `/function/system-metrics/figlet`

The 60m can be hard-coded and implicit. 

Response

```
{
  "success": 1005,
  "failure": 50
}
```

Ideally we should think about average execution latency, but this is a stretch goal.

ivanayov:

## Description

This function gets metricts from Prometheus for the number of
2xx and non 2xx status codes returned after function invocation.
It prints the result in json format like:
```
{
  success: 35
  failure: 3
}
```

Work done, but needs review and additional testing.

Resolves #162 

## How Has This Been Tested?

Invoked with request `http://127.0.0.1:8080/function/test-metrics`, where `test-metrics` was invoked several times, reproducing 2xx and non 2xx status codes.

## Checklist:

I have:

- [ ] updated the documentation if required - TODO
- [x] added unit tests


ivanayov:
Needs vendor update.
To simplify review, check only `expose-metrics/` files and `stack.yml`

expose-metrics/hadler_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/openfaas/faas/gateway/metrics"
+)
+
+func Test_parseFunctionName(t *testing.T) {
+
+	query := "http://127.0.0.1:8080/function/fnName"
```
alexellis: Named test scenarios would be more maintainable here. See Derek for examples.

expose-metrics/hadler_test.go
```diff
@@ -0,0 +1,82 @@
...
+	err := json.Unmarshal(val, &queryRes)
+	return &queryRes, err
+}
+
+func makeFakePrometheusQueryFetcher() FakePrometheusQueryFetcher {
+	return FakePrometheusQueryFetcher{}
+}
+
+func Test_getMetrics(t *testing.T) {
+	fakeQuery := makeFakePrometheusQueryFetcher()
```
alexellis: Would like to see this expanded. How are you treating 3xx?

expose-metrics/handler.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		log.Fatalf("Couldn't marshal json %t", err)
+	}
+	return string(res)
+}
+
+func parseFunctionName(query string) (string, error) {
+	var fnName string
+
+	matcher := regexp.MustCompile("^/?(?:/system/)??(?:async-)?function/([^/?]+)([^?]*)")
```
alexellis: What’s this for?

expose-metrics/handler.go
```diff
@@ -0,0 +1,111 @@
...
+
+	result := &Metrics{
+		Success: success,
+		Failure: failure,
+	}
+
+	return result, nil
+}
+
+func isSuccess(code string) bool {
```
alexellis: My comment re: 3xx statuses applies here too.

expose-metrics/handler.go
```diff
@@ -0,0 +1,111 @@
...
+	if code == strconv.Itoa(http.StatusOK) || code == strconv.Itoa(http.StatusAccepted) {
```
alexellis: Best not to Itoa twice. In fact these exist as string constants which is more performant for comparison.

stack.yml
```diff
@@ -188,3 +188,10 @@ functions:
       write_debug: true
       read_debug: true
 
+  expose-metrics:
+    lang: go
+    handler: ./expose-metrics
+    image: functions/expose-metrics:0.1.0
+    environment:
+      prometheus_host: prometheus
```
alexellis: This won’t work on Kubernetes (do you know why?) and should be placed in an external env file for consistency.

expose-metrics/handler.go
```diff
@@ -0,0 +1,111 @@
...
+	if len(matches) >= nameIndex+1 {
+		fnName = matches[nameIndex]
+	}
+
+	return fnName, nil
+}
+
+func getMetrics(fnName string, metricsQuery metrics.PrometheusQueryFetcher) (*Metrics, error) {
+
+	expr := url.QueryEscape(fmt.Sprintf(`sum(increase(gateway_function_invocation_total{function_name="%s"}[60m])) by (code)`, fnName))
```
stefanprodan: I would make the `60m` a parameter in this function that's configurable via an env var. The default value can be `60m` if no env var is supplied.

expose-metrics/handler.go
stefanprodan: Is would change this to `if code >= 200 && code < 400`

expose-metrics/handler.go
bartsmykla: Currently I can see that function is taking body as an url and parsing the function name from it. Can we assume that the body of a request will be just a function name instead of url?

expose-metrics/handler.go
bartsmykla: How would you name the env var? `metrics_time_period`, `metrics_interval`?

stack.yml
bartsmykla: @alexellis for kubernetes we need to specify namespace so the host would be `prometheus.openfaas` I think.

bartsmykla: @alexellis do you think that period of time for metrics (`60m` by default) should be configurable via env var as @stefanprodan suggested? If so, do you think that putting that env var with prometheus host/port is a good idea? Like this:
```
environment:
  prometheus_host: prometheus.openfaas
  prometheus_port: 9090
  metrics_time_period: 60m
```

expose-metrics/handler.go
bartsmykla: If so I think we could just get rid of that function and its tests

stack.yml
alexellis: Place it in `gateway_config.yml` please

expose-metrics/handler.go
alexellis: `metrics_window`

expose-metrics/handler.go
bartsmykla: If we'll assume that body of request == functionName, should I do some checking/sanitising of the body before proceeding with request?

alexellis: You should not accept a post body at all, this should be passed via HTTP path or query string.

alexellis: Also the body of the request would/should never contain /system /async or the word function? Where did we get that from?

bartsmykla:
@alexellis @stefanprodan I have created PR to @ivanayov branch: https://github.com/ivanayov/openfaas-cloud/pull/1 do you think that I should create separate PR with @ivanayov work instead?

alexellis:
This is what I do:

* Clone the remote repo
* Checkout their branch
* Type in `git checkout -b alexellis/feature`
* Make my changes and push the branch to my remote fork
* Raise a new PR and state "Closes #original_pr_number"

This way you have the commits & credits without any mess and you control the PR now.

Alex

alexellis:
Closed by 182

PR #194, opened 2018-09-19
Increase the default memory limit to 128 mb

martindekov:
Increasing the default memory limit of buildshiprun
in case the env-var is missing to be 128 mb

## Description

The default memory limit of buildshiprun was 20 mb in case the environmental variable is missing, now its 128 mb

## How Has This Been Tested?

N.A.

N.A.

## Checklist:

I have:

- [ ] added unit tests


martindekov:
Yeah I noticed that, thanks for the feedback!

PR #182, opened 2018-09-17
Feature/expose metrics

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/pull/177

## Description

This PR resolves comments from PR review: https://github.com/openfaas/openfaas-cloud/pull/177#issuecomment-422001435

## How Has This Been Tested?
1. Deployed `expose-metrics` to the cloud
2. Tested with
```
http://gw.iy.team-serverless.xyz:8080/function/expose-metrics?function=ivanayov-hello-wrold
```
Result
```json
{"success":4,"failure":2}
```

## Checklist:

I have:

- [ ] updated the documentation if required
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [x] added unit tests

PR #196, opened 2018-09-19
Add system-metrics README

ivanayov:
Describe what `system-metrics` function does, how to be invoked
together with some notes for function usage

Follows up #182

Refers to issue #162

### DIFF
--- a/system-metrics/README.md
+++ b/system-metrics/README.md
@@ -1,0 +1,20 @@
+## system-metrics
+
+This function exposes metrics from Prometheus.
+
+It queries Prometheus for a function's 2xx and non-2xx invocations count and returns a JSON response of type:
+
+```json
+{
+    "success": 10,
+    "failure": 8
+}
+```
+
+It takes function's name (i.e. `myFunction`) from a query and is invoked by GET request to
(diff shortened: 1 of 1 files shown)


system-metrics/README.md
```diff
@@ -0,0 +1,20 @@
...
+It queries Prometheus for a function's 2xx and non-2xx invocations count and returns a JSON response of type:
+
+```json
+{
+    "success": 10,
+    "failure": 8
+}
+```
+
+It takes function's name (f.e. `myFunction`) from a query and is invoked by GET request to
```
alexellis: f.e - is this meant to be i.e.?

system-metrics/README.md
```diff
@@ -0,0 +1,20 @@
...
+http://gateway-url:8080/function/system-metrics?function=myFunction
```
alexellis: Does it only work on query-string or HTTP path also?

ivanayov: For the moment it doesn't work with a path. Does this need an update?

alexellis: I think it's ok for now

Issue #145, opened 2018-09-07
Feature: OAuth2.0


alexellis:
According to the roadmap we should use an OAuth2.0 integration to personalise/protect the dashboard page.

## Current behaviour:

Dashboard pages are open

## New behaviour

User must first log-in with GitHub

I expect this to be implemented within the `router` component in a stateless way using the Code Grant flow from OAuth 2.0. Once authenticated the browser will be issued with a Cookie - most likely using JWT so that we can sign it with a key for future verification.

The router should redirect requests to unauthorised resources to the login page for a list of URLS such as:

* `/function/system-dashboard` (authorisation required)

alexellis:
I am making progress on this and will be posting an update soon.

alexellis:
Related: https://github.com/openfaas/openfaas-cloud/issues/192

martindekov:
Don't we use OAuth 2.0 for GitLab and GitHub now? If so is this closeable?

alexellis:
Derek close: feature released

PR #197, opened 2018-09-19
Add initial feature for auth proxy

alexellis:
## Description

- The router component now delegates auth to (auth) microservice
for each request. Protected routes need a cookie to proceed or are
asked to redirect to GitHub to obtain a cookie as part of the
authorization flow.

- Work remaining the auth service must sign the cookie ising a
private key and then validate it on each HTTP request using its
public key.

A GitHub app must be created in order to use the auth delegation,
this must be configured via environmental variables but will move
to secrets.

## Context

This is part of a feature to enable OAuth 2.0 authz via #145 - user dashboards should be private and not public/open. This PR is the beginning of that work and uses GitHub for "login".

## How Has This Been Tested?

Tested via `docker run` using the new Dockerfiles. An OpenFaaS gateway running on GKE was used for the router's upstream_url and a new GitHub OAuth app for the authz flow.

A gist was shared with around half a dozen users / community members who verified the flow.

## Remaining work

As per comment above.

~~The cookie is not signed and does not have "secure" turned on (HTTPS-only). The cookie also contains a (very limited) access-token issued by GitHub in plain-text, if you retrieve it from the Inspector page in your browser it has permissions to fetch your public profile.~~

~~The next steps are to sign the cookie as a JWT token and remove the access token or to encrypt it.~~

GitHub does not support OIDC so an `id_token` cannot be issued - instead this needs to be handled by the auth flow.

- [x] Sign JWT and store in cookie
- [x] Validate cookie with public key
- [x] Remove JSON struct from cookie leaving only JWT
- [ ] Update dashboard to read claims for the user id for dashboard 
- [x] Add any additional claims needed by overriding the "standard claims"

Once deployed all users will need to authenticate via their GitHub credentials to access the dashboard. 

I will need to add a work-around for developers so that they can turn off the auth delegation or override the list of protected routes.

## Architecture / conceptual-flow

oauth 2 0 for openfaas cloud

## Checklist:

I have:

- [x] added unit tests

## To test / validate

* Fetch from origin
* `git checkout alexellis/oauth2`
* Deploy OpenFaaS Cloud via the stack.yml file
* On Kubernetes deploy the yaml/core files
* Now follow the "Auth" steps in docs/README.md also creating a GitHub App
* Generate a key-pair for the auth service using the instructions found there
* Deploy the auth service
* Setup CNAME / or /etc/hosts file entries


auth/README.md
```diff
@@ -0,0 +1,59 @@
+auth
+=======
+
+The auth service can be used to evaluate whether access is available to a given resource
```
johnmccabe: Should resource be route?

johnmccabe:
@alexellis looks like the `template` dir is missing from `auth` (I've a review with more comments but wanted to flag this asap).

johnmccabe:
Yeah got it, thanks

auth/handlers/homepage.go
```diff
@@ -0,0 +1,67 @@
...
+	if keyErr != nil {
+		log.Fatal("Load public key", keyErr)
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println(r.URL.Path)
+
+		cookie, err := r.Cookie(cookieName)
+		if err != nil {
+			fmt.Println("No cookie found.")
```
johnmccabe: Is there a reason to jump between `fmt` and `log`?

auth/handlers/login.go
```diff
@@ -0,0 +1,36 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println("Path", r.URL.Path)
+
+		resource := "/"
+
+		if val := r.URL.Query().Get("r"); len(val) > 0 {
+			resource = val
+		}
+
+		if strings.EqualFold(r.URL.Path, "/login/github") {
```
johnmccabe: Does `/login/github/` need to be valid as well?

auth/handlers/oauth2.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	privateKey, keyErr := jwt.ParseECPrivateKeyFromPEM(privateKeydata)
+	if keyErr != nil {
+		log.Fatal("Load private key", keyErr)
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		log.Printf(`OAuth 2 - "%s"`, r.URL.Path)
+		if r.URL.Path != "/oauth2/authorized" {
```
johnmccabe: Does `/oauth2/authorized/` need to be checked?

auth/handlers/query.go
```diff
@@ -0,0 +1,101 @@
...
+
+		status := http.StatusOK
+		if len(resource) == 0 {
+			status = http.StatusBadRequest
+		} else if isProtected(resource, protected) {
+			started := time.Now()
+			cookieStatus := validCookie(r, cookieName, publicKey)
+			log.Printf("Cookie verified: %fs [%d]", time.Since(started).Seconds(), cookieStatus)
+
+			if cookieStatus == http.StatusOK {
```
johnmccabe: `switch` would be cleaner here.

auth/handlers/common.go
```diff
@@ -0,0 +1,56 @@
...
+	authURL := "https://github.com/login/oauth/authorize"
+	u, _ := url.Parse(authURL)
+	q := u.Query()
+
+	q.Set("scope", scope)
+	q.Set("allow_signup", "0")
+	q.Set("state", fmt.Sprintf("%d", time.Now().Unix()))
+	q.Set("client_id", config.ClientID)
+
+	redirectURI := combineURL(config.ExternalRedirectDomain, "/oauth2/authorized")
```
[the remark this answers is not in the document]
alexellis: What do you mean?

PRIVACY.md
```diff
@@ -20,17 +20,24 @@ Within the OpenFaaS API:
...
 
-No secret information is stored in OpenFaaS Cloud and minimal access is granted through OAuth permission scopes including: read/write to statuses and reading repository contents.
+No secret information from your GitHub profile is stored in OpenFaaS Cloud and minimal access is granted through OAuth permission scopes including: read/write to statuses and reading repository contents.
 
-Uninstalling
+### Authentication
 
-If you want to remove your endpoints from OpenFaaS Cloud please remove the GitHub app integration via your GitHub profile page and contact alex@openfaas.com to have your functions and container images removed from OpenFaaS Cloud.
+When the feature is live OpenFaaS Cloud will require a login to access the UI dashboard. Access to this dashboard will be via single sign-on (SSO) with your GitHub account through use of a GitHub OAuth 2.0 App. Once authenticated a JWT token will be issued to you containing claims about your user account - the JWT is stored in a Cookie in your browser, but is not stored on  OpenFaaS Cloud. If you wish to remove your cookie, just remove it from your browser or clear your browsing history.
```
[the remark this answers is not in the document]
alexellis: Good point, can you raise an issue for this UI feature?

auth/handlers/homepage.go
```diff
@@ -0,0 +1,67 @@
...
+// MakeHomepageHandler shows the homepage
+func MakeHomepageHandler(config *Config) func(http.ResponseWriter, *http.Request) {
+	keydata, err := ioutil.ReadFile(config.PublicKeyPath)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	publicKey, keyErr := jwt.ParseECPublicKeyFromPEM(keydata)
+	if keyErr != nil {
+		log.Fatal("Load public key", keyErr)
```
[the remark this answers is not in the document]
alexellis: I think it's gone too far to update at this point, but good for reference.

auth/handlers/oauth2.go
alexellis: I'm not following.

auth/handlers/login.go
alexellis: I'm missing something. `if strings.EqualFold(r.URL.Path, "/login/github")` this checks for the r.URL.Path to be exactly equal to `/login/github` right?

PRIVACY.md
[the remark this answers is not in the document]
alexellis: I believe the behaviour would be in the dashboard UI page itself.

auth/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+COPY static     static
+COPY template   template
+COPY provider   provider
+COPY main.go    .
+
+RUN go test -v \
+    && go build -o /usr/bin/auth .
+
+FROM alpine:3.8
+RUN apk add --no-cache ca-certificates
```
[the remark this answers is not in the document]
alexellis: We don't do this anywhere else

auth/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+For example:
+
+Check access to resource (r) /function/system-dashboard:
+
+```
```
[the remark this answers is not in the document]
alexellis: This is tedious to do for every quoted line

auth/handlers/common.go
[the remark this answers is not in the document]
fntlnz: he means that `path.Join` is the way to do what you did in `combineURL` provided by the standard library. Also `path.Join` is variadic.
https://play.golang.org/p/5BVOeHDCOay

auth/handlers/common.go
```diff
@@ -0,0 +1,56 @@
...
+
+	q.Set("redirect_uri", redirectURI)
+
+	u.RawQuery = q.Encode()
+	return u
+}
+
+func combineURL(a, b string) string {
```
fntlnz: just use `path.Join`

auth/handlers/customers.go
```diff
@@ -0,0 +1,124 @@
...
+	found := false
+
+	log.Printf("CUSTOMERS cache expires in: %fs", c.Expires.Sub(time.Now()).Seconds())
+	if c.Expires.Before(time.Now()) {
+		c.Fetch()
+	}
+
+	c.Sync.Lock()
+	lookup := *c.Usernames
+
```
fntlnz: If something goes wrong while doing the lookup the sync is never unlocked.

I would do something like:

```go
c.Sync.Lock()
defer c.Sync.Unlock()
lookup := *c.Usernames

if _, ok := lookup[strings.ToLower(login)]; ok {
	return true, nil
}
return false, nil

```

The benefit of this is also that any further change in this code is less likely to brak the unlock logic that is happening anyway.

auth/handlers/customers.go
```diff
@@ -0,0 +1,124 @@
...
+		log.Printf("unable to fetch customers from %s, error: %s", customersURL, getErr.Error())
+		return getErr
+	}
+
+	usernames := map[string]string{}
+	for _, customer := range customers {
+		usernames[customer] = "true"
+	}
+
+	c.Sync.Lock()
```
fntlnz: defer the unlock here too

auth/handlers/customers.go
```diff
@@ -0,0 +1,124 @@
...
+		for i, c := range customers {
+			customers[i] = strings.ToLower(c)
+		}
+
+	}
+
+	return customers, nil
+}
+
+// func validCustomer(customers []string, owner string) bool {
```
fntlnz: needed?

auth/handlers/customers_test.go
```diff
@@ -0,0 +1,14 @@
+package handlers
+
+import (
+	"log"
+	"testing"
+)
+
+func TestGet(t *testing.T) {
+	c := NewCustomers()
+	log.Println(c.Get("alexellis"))
```
fntlnz: assertions here?

auth/provider/github_profile.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return profile, fmt.Errorf("bad status code: %d", res.StatusCode)
+	}
+
+	if res.Body != nil {
+		defer res.Body.Close()
+
+		bytesOut, _ := ioutil.ReadAll(res.Body)
```
fntlnz: check error here

auth/handlers/customers_test.go
alexellis: Yes these tests need commenting out or having assertions adding.

auth/handlers/common.go
alexellis: Not sure it works the same.

alexellis:
@fntlnz @johnmccabe this playground example demonstrates why it's required to use `combineURL` - https://play.golang.org/p/NBknOKdQRei

```
https:/www.google.com/a/b
https://www.google.com/a/b
https:/www.google.com/a/b
https://www.google.com/a/b
```

Top version is from `path.Join`.

johnmccabe:
The example in my comment showed how to use path.Join, you'd be joining the .Path of the URL rather than the string. I'll share a playground example when I get to the office.

auth/README.md
```diff
@@ -0,0 +1,59 @@
...
+(JSON payload)
+
+| Field      | Description                     | Required |
+|------------|---------------------------------|----------|
+| sub        | unique ID for user              | true     |
+| login      | login in GitHub                 | true     |
+| full_name  | user's full-name as per profile | false    |
+
+
+## Building
```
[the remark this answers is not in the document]
alexellis: These are now covered in the README

PRIVACY.md
```diff
@@ -20,17 +20,24 @@ Within the OpenFaaS API:
...
+
+You may retract access to the GitHub OAuth 2.0 App by removing this at any time.
+
+### Uninstalling
+
+If you want to remove your endpoints from OpenFaaS Cloud please remove the GitHub App integration and GitHub OAuth 2.0 App via your GitHub profile page. This action will cause the deployed functions to be deleted, but the Docker images may still remain in the registry for a period of time after this step depending on whether a public Docker Hub account is being used or a private registry. To have all information removed contact alex@openfaas.com.
```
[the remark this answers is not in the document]
alexellis: You're right, but I think it's OK to be specific with this in the README. There is a difference between OAuth 1.0 vs 2.0

alexellis:
I'd take a quick PR for that @johnmccabe

PR #198, opened 2018-09-19
Format gateway url depending on env var

martindekov:
Creating a function in sdk which handles the formating of the
internal gateway URL. If variable present for kubernetes
the namespace is appended to the URL

## Description

The gateway will be formated depending the environmental variable `dns_suffix` placed in `gateway_config.yml` rather than manually go around all the places `.openfaas` for example is mentioned.

In code is used like 
```
  suffix := os.Getenv("dns_suffix")
  builderURL := os.Getenv("builder_url")
  builderURL = sdk.CreateServiceURL(builderURL, suffix)
  gatewayURL = sdk.CreateServiceURL(gatewayURL, suffix)
```

Once we have this re-vendored in all the functions I can move forward with updating the Documentation and all the functions.

## How Has This Been Tested?

Tested on Swarm and Kubernetes with one function `buildshiprun` and environmental variable `dns_suffix: openfaas` commented out(for Swarm) and non-commented out(for Kubernetes). The URL is built as it should and functions are built through `buildshiprun` as they should.

N.A. This feature wont affect anyone with instances of the cloud running.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/sdk/service.go
+++ b/sdk/service.go
@@ -8,3 +8,17 @@ import (
 func FormatServiceName(owner, functionName string) string {
 	return fmt.Sprintf("%s-%s", strings.ToLower(owner), functionName)
 }
+
+func CreateServiceURL(URL, suffix string) string {
+	if len(suffix) == 0 {
+		return URL
+	}
+	columns := strings.Count(URL, ":")
+	//columns in URL with port are 2 i.e. http://url:port
+	if columns == 2 {
+		baseURL := URL[:strings.LastIndex(URL, ":")]
+		port := URL[strings.LastIndex(URL, ":"):]
+		return fmt.Sprintf("%s.%s%s", baseURL, suffix, port)
+	}
+	return fmt.Sprintf("%s.%s", URL, suffix)
+}
--- a/sdk/service_test.go
+++ b/sdk/service_test.go
@@ -38,3 +38,45 @@ func Test_FormatServiceName(t *testing.T) {
 		}
 	}
 }
+
+func Test_CreateServiceURL(t *testing.T) {
+	tests := []struct {
+		title       string
+		URL         string
+		suffix      string
+		expectedURL string
+	}{
+		{
+			title:       "URL formatted for Swarm with port",
+			URL:         "http://gateway:8080",
+			suffix:      "",
+			expectedURL: "http://gateway:8080",
+		},
+		{
+			title:       "URL formatted for Kubernetes",
+			URL:         "http://gateway:8080",
+			suffix:      "openfaas",
+			expectedURL: "http://gateway.openfaas:8080",
+		},
+		{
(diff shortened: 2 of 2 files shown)


sdk/service.go
```diff
@@ -2,9 +2,20 @@ package sdk
...
 	"fmt"
+	"os"
 	"strings"
 )
 
 func FormatServiceName(owner, functionName string) string {
 	return fmt.Sprintf("%s-%s", strings.ToLower(owner), functionName)
 }
+
+func FormatGateway(URL string) string {
```
alexellis: This wouldn't be format `gateway` - it would be `createServiceURL()`. We should also allow the suffix to be passed in.

sdk/service_test.go
```diff
@@ -38,3 +39,34 @@ func Test_FormatServiceName(t *testing.T) {
...
+
+func Test_FormatGatewaySwarm(t *testing.T) {
+	tests := []struct {
+		title       string
+		URL         string
+		suffix      string
+		expectedURL string
+	}{
+		{
+			title:       "URL formatted for Swarm",
```
alexellis: Now do a test where there is no port i.e. `8080` is not given.

martindekov: It will fail if the port is not given.I though the format should always be with port for internal network?

sdk/service.go
martindekov: Yeah ok I will reformat that

sdk/service_test.go
alexellis: It won't fail if a port is not given, remember that we have more than one service we have to resolve, not just the gateway. If no port is given http:// means port 80 and https:// means port 443

sdk/service.go
alexellis: I also think we may be better returning `url.URL` than string https://golang.org/pkg/net/url/#URL

sdk/service.go
```diff
@@ -2,9 +2,20 @@ package sdk
...
+	suffix := os.Getenv("dns_suffix")
```
alexellis: This should be passed in.

sdk/service.go
martindekov: My first approach was with url.URL but I think its unnecessary complication. Just added simple if to resolve the port problem will amend now and check your feedback

martindekov:
Now this references with/out port with/out suffix but expects the URL to always start with `http://...` or `https://...`

martindekov:
In code is used like 
```
  suffix := os.Getenv("dns_suffix")
    builderURL := os.Getenv("builder_url")
    builderURL = sdk.CreateServiceURL(builderURL, suffix)
    gatewayURL = sdk.CreateServiceURL(gatewayURL, suffix)
```

Issue #191, opened 2018-09-18
UI Dashboard: Endpoint truncated to "Endpoi..."


alexellis:
## Expected Behaviour

"Endpoint" should show the full word, but doesn't

The icon for "Invocations" should not wrap onto a new line either - it looks a bit messy.

## Current Behaviour

The word "Endpoint" is is truncated to "Endpoi..."

## Possible Solution

I think we should change this to "Link:" but also implement the icons from #190 

![wrapping](https://user-images.githubusercontent.com/6358735/45709463-05a77700-bb7c-11e8-8f24-87466d618a12.png)

alexellis:
Derek add label: area/dashboard

bartsmykla:
Derek assign: me

Issue #190, opened 2018-09-18
UI Dashboard: change endpoint link to "copy to clipboard"


alexellis:
This change is for the Overview page of the dashboard.

## Expected Behaviour

The endpoint link may not be useful for many people to visit from the dashboard, so let's change the current behaviour of the link to "copy to clipboard" then add another link by it with the icon to show "open in new window"

https://fontawesome.com/icons/link?style=solid

https://fontawesome.com/icons/external-link-alt?style=solid

Suggestion: from Allen / Pato.

## Current Behaviour

The endpoint link will open the endpoint and "copy" is done by right-clicking and clicking "Copy link location"

Update dashboard/client, run `make` locally but do not commit the `dist` folder.

kenfdev:
Derek add label: area/dashboard

kenfdev:
FYI, the gateway UI implements it like this:

https://github.com/openfaas/faas/blob/master/gateway/assets/script/bootstrap.js#L26

Care is needed for browsers who support `document.queryCommandSupported('copy')`

kenfdev:
Derek add label: skill/beginner

alexellis:
Good comments Ken. This should be a quick PR for someone to pick up.

bartsmykla:
Derek assign: me

bartsmykla:
Estimate: 30 minut

bartsmykla:
![image](https://user-images.githubusercontent.com/11655498/45829323-9607c800-bcf2-11e8-8cc8-67a2661fd651.png)

alexellis:
Hi Bart, what is the image for?

This task is for the overview page (it may also apply to the details page) 

Alex

jaigouk:
@alexellis @bartsmykla 
I added `react-copy-to-clipboard` and the commit was not merged. so I am thinking about creating a new PR for the behaviour and then I found this ticket. 

![copy-badge-md](https://user-images.githubusercontent.com/6720/45879356-792bcd00-bda4-11e8-97c3-dbca5cddac7f.gif)

martindekov:
Derek close: we no longer support overview page and saw @bartsmykla tried to close this in a PR

PR #200, opened 2018-09-20
Small UI changes inside function dashboard

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/issues/190
Closes https://github.com/openfaas/openfaas-cloud/issues/191

Created CopyText component which after click will copy
some value to clipboard.

Invocation's icon will hide below some display size, so it
doesn't look messy anymore.

Endpoint is not being truncated because there is min-width
value set

## How Has This Been Tested?

These are small visual changes, so I have manually tested them on Chrome, Firefox and Safari.

There is no need to do anything special

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/dashboard/Makefile
+++ b/dashboard/Makefile
@@ -1,4 +1,4 @@
 .PHONY: build
 
 build:
-	( docker run -v $(shell pwd):/dashboard node:8-alpine /bin/sh -c "cd dashboard/client && yarn && yarn run build")
+	( docker run -v "$(shell pwd):/dashboard" node:8-alpine /bin/sh -c "cd dashboard/client && yarn && yarn run build")
--- a/dashboard/client/package.json
+++ b/dashboard/client/package.json
@@ -8,10 +8,12 @@
     "@fortawesome/react-fontawesome": "^0.1.2",
     "axios": "^0.18.0",
     "moment": "^2.22.2",
+    "prop-types": "^15.6.2",
     "query-string": "^5.1.1",
     "react": "^16.4.2",
     "react-ace": "^6.1.4",
     "react-dom": "^16.4.2",
+    "react-notifications": "^1.4.3",
     "react-router-dom": "^4.3.1",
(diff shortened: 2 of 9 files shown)


alexellis:
Do you have a screenshot or even better a gif?

bartsmykla:
@alexellis I'll do that

bartsmykla:
kapture 2018-09-21 at 9 56 05

alexellis:
Not sure that clicking the endpoint should copy to clipboard. The task is to add two FA icons one to open and one to copy. This also needs applying to the overview.

alexellis:
I'd rather see the user-feedback next to the button that was clicked than zooming in from the right. 

FYI: I'm seeing a bit of a rebase issue on this now too.

jagreehal:
@alexellis I'm not sure about having an icon to open a link, the hyperlink should suffice. I'd rather just have an icon to copy the link, which I think should be in a button.

@bartsmykla looks good. 

```
 document.execCommand('copy');
```

is [supported by a lot of browsers](https://caniuse.com/#feat=document-execcommand) and should be good enough to use... unless I've missed something.

Do we need such a prominent notification? Here's an example I came up with

example

The branch is here https://github.com/openfaas/openfaas-cloud/compare/master...jagreehal:issue-190?expand=1

Also for font awesome the existing icons are added to the 'library' in index.js. I'll have to check what's the best way to include them. I've gone for consitency but like your approach of pulling in what you need for the component.

jaigouk:
@jagreehal @bartsmykla When I worked on this PR https://github.com/openfaas/openfaas-cloud/pull/208, I found following links and ended up using `react-copy-to-clipboard`
https://stackoverflow.com/questions/51129888/document-execcommand-copy-dont-work-in-react
https://css-tricks.com/native-browser-copy-clipboard/

jaigouk:
@alexellis @kenfdev i will create a new PR to reflect the decision in badge modal.

Issue #172, opened 2018-09-13
UI: Dashboard GitHub badges


alexellis:
We should provide a button with a pop-up modal dialog that lets the user copy/paste markdown for a GitHub badge. This initial version won't connect to a live service but could just say `[OpenFaaS: Cloud]` for instance. Should be on the details page. See Travis CI for an example of this or the OpenFaaS readme.

alexellis:
Example:

![travis-web-example](https://user-images.githubusercontent.com/6358735/45512248-209e7380-b797-11e8-82c5-45f9f4496bc2.png)

jaigouk:
@alexellis i will start to work on this. if i can't finish it today then i will let you know.

alexellis:
Additional example:

```
[![OpenFaaS](https://img.shields.io/badge/openfaas-cloud-blue.svg)](https://www.openfaas.com)
```

[![OpenFaaS](https://img.shields.io/badge/openfaas-cloud-blue.svg)](https://www.openfaas.com)

Initially the pop-up would just give text as above, but in time we can extend it to potentially give the build status and other data like invocations.

alexellis:
It'd be good to see this modal after clicking "Get badge" or similar in the style of "Build logs".

jaigouk:
@kenfdev recommended to start using https://github.com/reactstrap/reactstrap

kenfdev:
I'd say we add `react-bootstrap` and just use the [Modal component](https://react-bootstrap.github.io/components/modal/) to show modals. The inclusion of `react-bootstrap` is optimized so the components we include will only be imported to the final bundled JavaScript.

I think this is the most minimal impact approach. I'd like Jag's input, too.

alexellis:
Please could you do a PR/PoC for this?

jaigouk:
would this enough for the badge?
<img width="938" alt="screen shot 2018-09-20 at 22 44 42" src="https://user-images.githubusercontent.com/6720/45846420-0250ee80-bd28-11e8-867b-dcdc58042c57.png">

alexellis:
Looks like a nice start, have you tried Ken's suggestion? Maybe you could try adding it and show a gif?

jaigouk:
added react-bootstrap "^0.32.4"

![get-badge](https://user-images.githubusercontent.com/6720/45847823-9624b980-bd2c-11e8-845e-b6770a64b1c1.gif)

alexellis:
[...]

PR #208, opened 2018-09-21
copy badge markdown to clipboard

jaigouk:
Signed-off-by: Jaigouk Kim <ping@jaigouk.kim>

## Description

- added react-copy-to-clipboard
- clicking well area in modal will copy the markdown
- change text input to "well" : github shows token in similar way. no text input.
- fixes 1 left over issue in #172 (copy button)

## How Has This Been Tested?

manually

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/dashboard/client/package.json
+++ b/dashboard/client/package.json
@@ -12,6 +12,7 @@
     "react": "^16.4.2",
     "react-ace": "^6.1.4",
     "react-bootstrap": "^0.32.4",
+    "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.4.2",
     "react-router-dom": "^4.3.1",
     "react-scripts": "1.1.5"
--- a/dashboard/client/src/pages/FunctionDetailPage.js
+++ b/dashboard/client/src/pages/FunctionDetailPage.js
@@ -1,10 +1,11 @@
 import React, { Component } from 'react';
 import queryString from 'query-string';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Link } from 'react-router-dom';
+import { Modal, Form, Well } from 'react-bootstrap';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
 import { functionsApi } from '../api/functionsApi';
 import { FunctionDetailSummary } from '../components/FunctionDetailSummary';
-import { Modal, Button, FormGroup, ControlLabel, FormControl } from 'react-bootstrap';
+
 
 export class FunctionDetailPage extends Component {
   constructor(props) {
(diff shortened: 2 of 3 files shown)


jaigouk:
need to rebase

jaigouk:
initially, for the badge markdown, I used input field. and then I found that it’s not ideal because if I put my cursor in it, it tried to show my past input texts in chrome. so i think it’s better to use `Well` like thingy.

I added `react-copy-to-clipboard` after researching. it seems like it can get messy with clipboard thingy. so i ended up using that lib. If there is a better way to do it, then I will try to adapt to the better solution for it. since the package is quite old.

alexellis:
Thanks for the update. @kenfdev what are your thoughts?

kenfdev:
I like the `well` approach. But I think it's better without the `copy` button. How about:

* just showing the `well`
* hovering above it changes the cursor to `pointer`
* click the `well` and try to copy to clipboard
* If copied, show `Copied` with a `tooltip`
* If copying is not supported, just select all of the text inside the `well`

WDYT?

I'm not sure about adding `react-copy-to-clipboard` as a dependency. What trouble did you have @jaigouk ? I know that it's a bit tricky to support all browsers, but I think we could leave those legacy ones.

jaigouk:
https://stackoverflow.com/questions/51129888/document-execcommand-copy-dont-work-in-react
https://css-tricks.com/native-browser-copy-clipboard/
there are some issues with using `document.execCommand ('copy')`

alexellis:
@kenfdev how are we looking now?

kenfdev:
@jaigouk This is looking awesome! Thanks a lot. One last opinion, the tooltip looks rather strange after having a look (sorry). How about just showing a "Copied!" text in green (or some different color) Can you also select all the text inside the well (so that the user can also do ctrl+c to copy)? I think this is a common behavior for these types of elements.

copy_badge_markdown_to_clipboard_by_jaigouk_ _pull_request__208_ _openfaas_openfaas-cloud

If this is complete, I think it'll be a great experience for the user.

jaigouk:
@kenfdev i checked github's token copy page. they don't select text. and same for aws console. 
5th-attempt-for-copying-badge
I think selecting text is not a big deal if users want to do it.

jaigouk:
The most common pattern for this we see everyday is github's clone modal.
screen shot 2018-09-23 at 21 04 07

alexellis:
@jaigouk thanks for working through the feedback and collaborating with the community. I'll merge and we can make any additional tweaks from there. Looks great :medal_sports:

kenfdev:
@jaigouk Thank you very much for your thorough research and contribution.  This looks great 🎉

jaigouk:
@kenfdev thanks. let's see how it goes and tweak this based on users' feedback

PR #235, opened 2018-09-26
WIP: Added private repo label for private functions

bartsmykla:
If function was build from private github repository it will receive
label com.openfaas.cloud.git-private-repo=true

## How Has This Been Tested?
WIP

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -154,13 +154,14 @@ func Handle(req []byte) string {
 			Image:   imageName,
 			Network: "func_functions",
 			Labels: map[string]string{
-				sdk.FunctionLabelPrefix + "git-cloud":      "1",
-				sdk.FunctionLabelPrefix + "git-owner":      event.Owner,
-				sdk.FunctionLabelPrefix + "git-repo":       event.Repository,
-				sdk.FunctionLabelPrefix + "git-deploytime": strconv.FormatInt(time.Now().Unix(), 10), //Unix Epoch string
-				sdk.FunctionLabelPrefix + "git-sha":        event.SHA,
-				"faas_function":                            serviceValue,
-				"app":                                      serviceValue,
+				sdk.FunctionLabelPrefix + "git-cloud":        "1",
+				sdk.FunctionLabelPrefix + "git-owner":        event.Owner,
+				sdk.FunctionLabelPrefix + "git-repo":         event.Repository,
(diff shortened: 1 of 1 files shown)


alexellis:
Why is this WIP (please say)? Do we have any private repositories deployed? (That would be an impact to the users)

bartsmykla:
It's WIP, because it's not working. It's assuming that `event` inside builldsiprun will have information if repo is private or not, but inside `getEvent` function there is no notion if the repository is private, so I was looking at env vars which can tell if it's private or not, but without success for now.

buildshiprun/handler.go
```diff
@@ -154,13 +154,14 @@ func Handle(req []byte) string {
...
-				sdk.FunctionLabelPrefix + "git-deploytime": strconv.FormatInt(time.Now().Unix(), 10), //Unix Epoch string
-				sdk.FunctionLabelPrefix + "git-sha":        event.SHA,
-				"faas_function":                            serviceValue,
-				"app":                                      serviceValue,
+				sdk.FunctionLabelPrefix + "git-cloud":        "1",
+				sdk.FunctionLabelPrefix + "git-owner":        event.Owner,
+				sdk.FunctionLabelPrefix + "git-repo":         event.Repository,
+				sdk.FunctionLabelPrefix + "git-deploytime":   strconv.FormatInt(time.Now().Unix(), 10), //Unix Epoch string
+				sdk.FunctionLabelPrefix + "git-sha":          event.SHA,
+				sdk.FunctionLabelPrefix + "git-private-repo": strconv.FormatBool(event.Private),
```
alexellis: Is this data being populated by `git-tar` yet?

alexellis: cc @martindekov

martindekov: No It is not populated. I can handle the population of the struct in different PR and this one to extend it? But the answer where should we take whether or not is private or not is from payload ```...
  "repository": {
...
    "private": false,
...
``` Like pointed in the issue

alexellis:
This PR is simply rendering the data in the labels (if it's available) - we need another PR to populate this data so that this function can pick it up. Does that make sense?

You absolutely have to obtain the data from `github-push`.

buildshiprun/handler.go
alexellis: Also we used 0/1 elsewhere, so I think we should do the same here and not use "true/false"

PR #244, opened 2018-09-28
Call garbage-collect through async route

### LINKED ISSUE #221: Asynchronous invocation of garbage-collect
Currently in `github-event` we invoke `garbage-collect` in the synchronous route e.g. `/function/garbage-collect` which time outs in case we delete the whole application and we have couple of repositories installed on the cloud.

## Expected Behaviour

We should call garbage-collect from the asynchronous route e.g. `/async-function/garbage-collect` to handle the case where we uninstall the whole openfaas-cloud application and we have more then one repositories.

## Current Behaviour

We call the function in the synchronous route e.g. `/function/garbage-collect` and it will timeout if we have couple of repositories

## Possible Solution

Invoke garbage-collect from the asynchronous route

1.Install OF-cloud application on couple of repositories
2.Wait for your functions to be deployed
3.Uninstall whole app
4.See you still have functions on the cloud

## Context

For the times a user would like to remove his functions from the cloud and we want to automate the process.

## Your Environment

18.06.0-ce
Docker Swarm
Linux
N.A. try with public cloud if you have access
* Please also follow the [troubleshooting guide](https://github.com/openfaas/faas/blob/master/guide/troubleshooting.md) and paste in any other diagnostic information you have:

martindekov:
Calling garbage-collect through async route in github-event
so the request wont time out in case we have couple of functions
deployed

## Description

This is small tweak that enables calling garbage-collect through async route for the times user with couple of functions want to uninstall the application and delete all its functions from the cloud.

Closes #221 

## How Has This Been Tested?

All function deleted with logs logs 
```
...
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Need to remove: martindekov/bottester.
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Need to remove: martindekov/goforfun.
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Need to remove: martindekov/gocloudfunc.
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Need to remove: martindekov/cloudlogo.
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Need to remove: martindekov/kubernetesgo.
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Status code for garbageCollect: 202
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Status code for garbageCollect: 202
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Status code for garbageCollect: 202
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Status code for garbageCollect: 202
system-github-event.1.w4bvfwjy9l3l@FaasSwarm    | Status code for garbageCollect: 202
...
```

No migration needed, just added feature.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/github-event/handler.go
+++ b/github-event/handler.go
@@ -156,7 +156,7 @@ func garbageCollect(garbageRequests []GarbageRequest) error {
 
 		body, _ := json.Marshal(garbageRequest)
 		bodyReader := bytes.NewReader(body)
-		req, _ := http.NewRequest(http.MethodPost, gatewayURL+"function/garbage-collect", bodyReader)
+		req, _ := http.NewRequest(http.MethodPost, gatewayURL+"async-function/garbage-collect", bodyReader)
 
 		digest := hmac.Sign(body, []byte(payloadSecret))
 		req.Header.Add(sdk.CloudSignatureHeader, "sha1="+hex.EncodeToString(digest))
@@ -168,7 +168,8 @@ func garbageCollect(garbageRequests []GarbageRequest) error {
 		if res.Body != nil {
 			defer res.Body.Close()
 		}
-		if res.StatusCode != http.StatusOK {
+		if res.StatusCode != http.StatusAccepted {
+			log.Printf("Unexpected status code for function: `%s` - %d\n", garbageRequest.Repo, res.StatusCode)
 			resBody, _ := ioutil.ReadAll(res.Body)
 			fmt.Printf("Error in garbageCollect: %s\n", resBody)
 		}


alexellis:
```
		if res.StatusCode != http.StatusOK {
```

We should check for 202 in this instance, StatusAccepted (only)

alexellis:
Perhaps we should also include the status code in the message - this would have helped you in this instance.

github-event/handler.go
```diff
@@ -168,7 +168,8 @@ func garbageCollect(garbageRequests []GarbageRequest) error {
 		if res.Body != nil {
 			defer res.Body.Close()
 		}
-		if res.StatusCode != http.StatusOK {
+		fmt.Printf("Status code for garbageCollect: %d\n", res.StatusCode)
```
alexellis: Only `log.Printf` if this is not the desired status code?

martindekov: This the status code for every request printed since you had comment above that its good idea to print all the status codes to check whats up with every request `Perhaps we should also include the status code in the message - this would have helped you in this instance.`  I think that's what you meant.

ivanayov: You can add 
`garbageRequest.Owner/Repo` so that the message can point to the GC-ed function

martindekov:
Referenced your comment @ivanayov . Now I give repo - status code.

PR #250, opened 2018-09-28
Add caching for templates

### LINKED ISSUE #215: git-tar - cache templates
## Expected Behaviour

Given that templates rarely change, we should cache the 3-4 template repos we pull in from GitHub and copy them to the working directory for the shrink-wrap stage

## Current Behaviour

We clone all 3-4 repos for every function in the stack file on every `git push`. It's not necessary. I think we could cache to a known-folder in `/tmp/` and copy the files in from there.

martindekov:
Adding caching templates through two functions
one formats the clonepath in `/tmp` and the other
copies `templates` from `/tmp` to clonePath e.g.
`/tmp/templates` to `/tmp/username/function/` achieved
through changing the fetechTemplates function and
adding copyTemplates function

## Description

`fetchTemplates` function now fetch templates in `/tmp` rather than the clone path(`/tmp/username/function`) and added function `copyTemplates` which copies templates to the `clonePath`

Closes #215

## How Has This Been Tested?

Deployed two functions from two different accounts the output of git-tar is 

```
...
git-tar.1.j2ff8k9txc7a@FaasSwarm    | 2018/09/28 14:32:12 Functions owned by martindekov:
git-tar.1.j2ff8k9txc7a@FaasSwarm    |  [{martindekov-gocloudfunc martindekov-gocloudfunc-gocloudfunc:latest-50dc175 map[app:martindekov-gocloudfunc com.openfaas.cloud.git-cloud:1 com.openfaas.cloud.git-owner:martindekov com.openfaas.cloud.git-scm:github com.openfaas.cloud.git-sha:50dc17546057c41f69283496f78bd5b11fbd18ec com.openfaas.function:martindekov-gocloudfunc faas_function:martindekov-gocloudfunc com.openfaas.cloud.git-deploytime:1538145092 com.openfaas.cloud.git-private:0 com.openfaas.cloud.git-repo:gocloudfunc com.openfaas.scale.max:4 com.openfaas.scale.min:1 function:true]}]Garbage collection ran for martindekov/gocloudfunc - 0 functions deleted.
...
```
and
```
...
git-tar.1.j2ff8k9txc7a@FaasSwarm    |  [{martindekovwat-goforfun martindekovwat-exmpl-goforfun:latest-a4d1764 map[com.openfaas.cloud.git-deploytime:1538145248 com.openfaas.cloud.git-owner:MartinDekovWat com.openfaas.cloud.git-private:0 com.openfaas.scale.min:1 function:true faas_function:martindekovwat-goforfun app:martindekovwat-goforfun com.openfaas.cloud.git-cloud:1 com.openfaas.cloud.git-repo:exmpl com.openfaas.cloud.git-scm:github com.openfaas.cloud.git-sha:a4d17647b3e34daf41d5ac9860b6f429133e1cf4 com.openfaas.function:martindekovwat-goforfun com.openfaas.scale.max:4]}]Garbage collection ran for MartinDekovWat/exmpl - 0 functions deleted.
...
```

Content inside container of `git-tar` is 

```
/ $ cd /tmp
/tmp $ ls
MartinDekovWat  martindekov     template
```

Tested parallel fetching of URLs and the following massage appeared (because everything passed as expected) in logs
```
...
git-tar.1.2oh9d4wreof5@FaasSwarm    | Successfully fetched all templates.
...
```

Have the change in image `martindekov/of-git-tar:215` for testing

N.A. None.

## Checklist:

I have:

- [x] added unit tests


git-tar/function/handler.go
```diff
@@ -219,3 +229,22 @@ func reportStatus(status *sdk.Status) {
...
 	}
 }
+
+func formatTemplatePath(clonePath string) string {
+	folders := strings.Split(clonePath, "/")
+	return fmt.Sprintf("/%s", folders[1])
+}
+
+func copyTemplates(templatePath, clonePath string) error {
+	copyCmd := exec.Command("cp", "-r", "templates", clonePath)
```
alexellis: Please do not execute `cp`. See the approach we take in the faas-cli. If in doubt you can ask @rgee0 and @LucasRoesler about this since they have worked on.

git-tar/function/ops.go
```diff
@@ -43,21 +43,22 @@ func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error
 }
 
 func fetchTemplates(filePath string) error {
-	templateRepos := []string{"https://github.com/openfaas/templates", "https://github.com/openfaas-incubator/node8-express-template.git", "https://github.com/openfaas-incubator/golang-http-template.git"}
+	if _, err := os.Stat(filePath + "/template"); os.IsNotExist(err) {
+		templateRepos := []string{"https://github.com/openfaas/templates", "https://github.com/openfaas-incubator/node8-express-template.git", "https://github.com/openfaas-incubator/golang-http-template.git"}
```
alexellis: Could you use an error group for this instead so that they are executed in parallel? Look at the way this was done by @AkihiroSuda in the of-builder component.

git-tar/function/handler.go
martindekov: Their approach was my formal one but we discussed and Rosti told that `cp` uses some linux optimization.I will remake it how its done with `faas-cli`.

git-tar/function/ops.go
martindekov: Referenced that but I don't know why it wont make that comment outdated

git-tar/function/ops.go
```diff
@@ -43,18 +44,28 @@ func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error
...
-		pullCmd := exec.Command("faas-cli", "template", "pull", repo)
-		pullCmd.Dir = filePath
-		err := pullCmd.Start()
-		if err != nil {
-			return fmt.Errorf("Failed to start faas-cli template pull: %t", err)
+	if _, err := os.Stat(filePath + "/template"); os.IsNotExist(err) {
+		var eg errgroup.Group
+		templateRepos := []string{"https://github.com/openfaas/templates", "https://github.com/openfaas-incubator/node8-express-template.git", "https://github.com/openfaas-incubator/golang-http-template.git"}
+
+		for _, repo := range templateRepos {
```
alexellis: @AkihiroSuda please could you review this use of errgroup?

git-tar/function/ops.go
```diff
@@ -461,3 +472,76 @@ func getShortSHA(sha string) string {
...
+			filepath.Join(dest, info.Name()),
+		); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func copyFile(src, dest string) error {
```
alexellis: Can the package be vendored from the faas-cli, or does it need to be duplicated and modified? cc @LucasRoesler @rgee0

git-tar/function/ops.go
AkihiroSuda: Almost looks good but why ignore non-nil err from eg.Wait?

martindekov: Ah ... Thanks for the feedback @AkihiroSuda I forgot I missed the bad error case ..

git-tar/function/ops.go
martindekov: I totally stole the approach from you guys @rgee0 @LucasRoesler I think it should be vendored, but whatever you suggest I will go with it.

LucasRoesler: The original code in the faas-cli was meant to be generic and reusable, if we can reuse it here, all the better

martindekov:
Ok I have re-vendored `builder` from faas-cli

martindekov:
With custom templates feature I need to add logic to check the content of the environmental file and check the content of the template in folder. Will start working on that

git-tar/function/ops.go
```diff
@@ -44,18 +45,28 @@ func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error
...
 func fetchTemplates(filePath string) error {
-	templateRepos := formatTemplateRepos()
-
-	for _, repo := range templateRepos {
-		pullCmd := exec.Command("faas-cli", "template", "pull", repo)
-		pullCmd.Dir = filePath
-		err := pullCmd.Start()
-		if err != nil {
-			return fmt.Errorf("Failed to start faas-cli template pull: %t", err)
+	if _, err := os.Stat(filePath + "/template"); os.IsNotExist(err) {
```
martindekov: I had to do a review on my own code, since I forgot how it exactly worked, I guess now with the custom templates feature, things get more complicated as I need to know if the template is already downloaded, what if we want to overwrite/update the existing templates, should there be environmental variable disabling that ?

martindekov:
Hello there I would like some imput, ideas on how to improve this since when I started it the support was only for official templates, and not for custom ones, and what about the times you would like to update/overwrite the templates? Thanks for the input in advance!

cc @openfaas/faas-team

martindekov:
The ideas I have ATM, are adding environmental variable to enable/disable updating of the repositories(`--overwrite`), along with adding some function which trims the custom template url `https://github.com/openfaas-incubator/node8-express-template` -> `node8-express-template` and see if we have it ? 

This brings me to a question about the CLI as we input more and more custom templates(who knows?) maybe names will overlap(`template store openfaas/go`).

LucasRoesler:
I would just `git clone` the templates in `tmp` and the if the folder already exists, use `git pull`.  This will always update the templates when they change, but will reduce the bandwidth on most builds because the local copy will be up-to-date.  

I would follow the patten that `golang` uses in the `GOPATH`, the checkout of `github.com/openfaas-incubator/node8-express-template` should go to `/tmp/github.com/openfaas-incubator/node8-express-template`.  This avoids naming collisions since the repo address needs to be unique anyway. 

I am not sure a new flag  like `--overwrite` is needed.  What are the cases were you wouldn't want to update the template? Also how do you ensure that this function doesn't scale up and pull a new copy of the template anyway?  The template cache is ephemeral and local to the function

martindekov:
Hey @LucasRoesler thank you for the feedback on this!

> I am not sure a new flag like --overwrite is needed. What are the cases were you wouldn't want to update the template? Also how do you ensure that this function doesn't scale up and pull a new copy of the template anyway? The template cache is ephemeral and local to the function

Haven't really though of this scenario, when I opened the PR was still learning but in that case caching is not optimal since we scale the function its memory is deleted the moment we return to 1 replica ... along with the fact that the container may fail and be redeployed it renders this PR useless(caching in a container as a whole a bad idea?)

The whole point was copying the templates in a known location when they are first pulled instead of pulling all the templates every time the function is built, currently we do that and it is not who knows what for building time, but still and optimisation is generally a good idea I think.

I think that pulling templates in file structure like `go` does is generally a good idea(maybe slower search?) but Issue should be opened for faas-cli(WDYT?) since it is not connected with this change.

LucasRoesler:
I think there is some value in having a cache in the CLI

martindekov:
@alexellis Please could you look and give your opinion on this? Thanks

alexellis:
I was pretty sure we already have the overwrite flag. We don't clone a git repo because this is the approach taken by vendoring in Go and that's how Lucas implemented it. 

What's the confusion/disconnect? Basically the git-tar function should keep a cache of the templates as they are used so that we don't waste bandwidth.

martindekov:
What if the container scales, then when scaled to 5(example) replicas there won't have the cached templates and either way they will get pulled.

With the custom template logic there comes a problem, before caching we need to traverse the directory to check if the custom template is already there and if not it should get pulled along with the official templates if they are not pulled, wouldn't that slow down the things ?

martindekov:
The PR was opened when custom template was not though of and we only pulled the official templates

alexellis:
I see no problems in maintaining a cache of the templates from GitHub. What is the issue as you understand it?

martindekov:
Wouldn't traversing a directory every time we pull slow things down, and the benefit of optimization be minimal? If proceed I can do the following:

check if `template` folder exist in `/tmp` if so, check if custom templates exist, take the naming of the template from the URL check if the folder has the template if so skip it altogether if not pull the custom template

alexellis:
What is slower? Pulling several MBs over the internet, or checking if a folder exists on the disk? 

Remember that you can also create a file when you pull to show that it has been pulled. There should be no need to traverse a deep folder structure to work this out.

LucasRoesler:
The template name should uniquely identify a folder in the cache, so there shouldn't be a need to traverse it.  It should be a quick check using `os.Stat(path)` 

You have mentioned the custom templates a couple times, why is caching those an issue?

alexellis:
Closing as inactive, if we still want to do this work, let's raise a new issue to discuss. Thanks for the contribution.

PR #251, opened 2018-09-28
Add truncate method to github-status

martindekov:
Adding truncate method in github status

## Description

Truncate method to trim strings if they exceed given length e.g. "some string" max chars 5 will give "tring"

## How Has This Been Tested?

Created test for it

N.A.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/github-status/handler.go
+++ b/github-status/handler.go
@@ -328,3 +328,12 @@ func hmacEnabled() bool {
 func buildStatus(status string, desc string, context string, url string) *github.RepoStatus {
 	return &github.RepoStatus{State: &status, TargetURL: &url, Description: &desc, Context: &context}
 }
+
+func truncate(maxLength int, message string) (string, bool) {
+	truncated := false
+	if len(message) > maxLength {
+		message = message[len(message)-maxLength:]
+		truncated = true
+	}
+	return message, truncated
+}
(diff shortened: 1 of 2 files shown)


github-status/handler.go
```diff
@@ -328,3 +328,12 @@ func hmacEnabled() bool {
 func buildStatus(status string, desc string, context string, url string) *github.RepoStatus {
 	return &github.RepoStatus{State: &status, TargetURL: &url, Description: &desc, Context: &context}
 }
+
+func truncate(maxLength int, message string) (string, bool) {
+	truncated := false
+	if maxLength < len(message) {
```
alexellis: Write it the other way as this looks a bit odd

github-status/handler.go
```diff
@@ -328,3 +328,12 @@ func hmacEnabled() bool {
 func buildStatus(status string, desc string, context string, url string) *github.RepoStatus {
 	return &github.RepoStatus{State: &status, TargetURL: &url, Description: &desc, Context: &context}
 }
+
+func truncate(maxLength int, message string) (string, bool) {
```
alexellis: Can you connect this to the output of the get logs function so that we don't post too much data to github?

martindekov: What is the upper limit?

alexellis: The maximum size is 64 KB.

https://developer.github.com/v3/checks/runs/#create-a-check-run

PR #261, opened 2018-10-01
Add auditing for garbage-collect

### LINKED ISSUE #248: Add auditing for garbage-collect
## Expected Behaviour

We should get an audit report of any functions by name / count that are removed via garbage-collect.

## Current Behaviour

Available in container logs, but not in audit log

## Possible Solution

See approach used in github-push and update garbage-collect function.

martindekov:
Sending events to through auditEvent in the cases when:
 - deletion of funtions starts
 - function is not deleted/unable to delete
 - deletion of functions is finished

## Description

Added auditing for the `garbage-collect` function in three places:
 - deletion of funtions starts
 - function is not deleted/unable to delete
 - deletion of functions is finished

Closes #248 

Note: Did not add audit for all errors since the issue suggests that we need auditing around the removal process. I can add auditing on every error if requested.

## How Has This Been Tested?

N.A. Does not alter the main `garbage-collect` functionality.

N.A. This just audits the event it does not intercept the functionality

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/garbage-collect/handler.go
+++ b/garbage-collect/handler.go
@@ -15,6 +15,8 @@ import (
 	"github.com/openfaas/openfaas-cloud/sdk"
 )
 
+const Source = "garbage-collect"
+
 // Handle a serverless request
 func Handle(req []byte) string {
 	validateErr := validateRequest(req)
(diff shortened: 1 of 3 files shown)


ivanayov:
LGTM

garbage-collect/handler.go
```diff
@@ -42,6 +44,12 @@ func Handle(req []byte) string {
 		log.Fatal(err)
 	}
 
+	auditEvent := sdk.AuditEvent{
```
alexellis: I think we could avoid this pre-delete message. It may end-up being too spammy.

martindekov: Ok I will remove this

martindekov:
Referenced your comment @alexellis

Issue #163, opened 2018-09-12
Move CUSTOMERS validation out of github-push


martindekov:
Soon we would like to move out of `github-push` for event handling and support more than push events from github and give that to `github-event`. Currently the CUSTOMER validation is handled in `github-push`

## Expected Behaviour

Should extend the `PushEvent` struct with `sender` field to compare by that since the field is present in the outer layer of `X-GitHub-Event: push`,`X-GitHub-Event: installation_repositories` and `X-GitHub-Event: integration_installation_repositories`.

## Current Behaviour

We handle the situation in `github-push` by struct which is unique to the push event.

## Possible Solution

Move the CUSTOMERS handle out of github-push into github-event and create separate struct or extend existing struct which handles events so it can compare by `event.Sender.Owner` rather than more specific `event.PushEventRepository.Owner.Login` since every event contains
```
...
    "sender": {
        "login": "martindekov",
...
```

## Steps to Reproduce (for bugs)

N.A. 

## Context

If we want the `github-event` function to handle the redirection we cant contain CUSTOMERS validation in `github-push`

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):

18.06.0-ce

* Are you using Docker Swarm or Kubernetes (FaaS-netes)?

Docker Swarm

* Operating System and version (e.g. Linux, Windows, MacOS):

Ubuntu 18.04

* Link to your project or a code example to reproduce issue:

N.A.

martindekov:
I can work on that when we settle on solution

alexellis:
Sounds fine to me, but we have to switch over to github-event first and github-push must validate only with the internal trust signature.

alexellis:
@martindekov you can work on this now that I've merged the work to switch to github-event as the primary endpoint.

martindekov:
Ok I am on it

martindekov:
Derek assign: me

PR #271, opened 2018-10-08
Add Customer struct

martindekov:
Adding Customer struct which contains Login field to handle
the general case when user sends events to the cloud

## Description

Adding `Customer` struct for the owner of the event to be recognized in event rather than `github-push`

This is 1/2 Part of two part PR that referenced #163 . 

The struct is separate since it is present in every event we support.

## How Has This Been Tested?

N.A.

N.A. This is struct update

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/sdk/events.go
+++ b/sdk/events.go
@@ -41,6 +41,14 @@ type Event struct {
 	Private        bool              `json:"private"`
 }
 
+type Customer struct {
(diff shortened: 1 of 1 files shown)


ivanayov:
Given the struct follows the GitHub API, it looks good

alexellis:
Merged

PR #272, opened 2018-10-08
Move customer validation into github-event

### LINKED ISSUE #163: Move CUSTOMERS validation out of github-push

martindekov:
Moving customer validation out of github-push
and into github-event depending on commit #271

## Description

This commit extends #271 and if merged closes #163. Currently the build fails since the struct is not vendored.

> Note: Not sure if the `Customers` struct in #271 would be used elsewhere. Should I add it into the handler directly? 

Also added additional check for `\r` since if making changes to a file through github UI it will append `\r\n` to the end of the line rather than only `\n` as expected. I personally have separate repository that contains my customers for faster updating.

## How Has This Been Tested?

Have the change in image `martindekov/github-event:163` and deployed image on my instance.

Added my own Customers file and tested with case sensitive name existing in the file
```
Push - {refs/heads/master {exmpl MartinDekovWat/exmpl https://github.com/MartinDekovWat/exmpl.git false {MartinDekovWat 42210992+MartinDekovWat@users.noreply.github.com}} d86e067fc83d4ada19ce05cdf5edc141d35d0076 {312235}}, git-tar status: 202
```
And my account non existing in file
```
Customer: martindekov not found in CUSTOMERS file via https://raw.githubusercontent.com/martindekov/mycustomers/master/CUSTOMERS
```

N.A. moving logic out of the `github-push`

## Checklist:

I have:

- [ ] added unit tests


ivanayov:
The build fails with `function/handler.go:51:15: undefined: sdk.Customer`
You need to take #271

alexellis:
@martindekov why does this fail the CI?

martindekov:
Because this extends #271 once merged I will vendor the SDK and it wont fail.

martindekov:
Ah ok I will

ivanayov:
Looks good

alexellis:
Looks ready to merge.

github-event/handler.go
```diff
@@ -238,3 +268,56 @@ func forward(req []byte, function string, headers map[string]string) (string, in
...
+			break
+		}
+	}
+
+	return found
+}
+
+// getCustomers reads a list of customers separated by new lines
+// who are valid users of OpenFaaS cloud
+func getCustomers(customerURL string) ([]string, error) {
```
alexellis: How is this different from what we do for Derek? Can you take a look and compare?

github-event/handler.go
```diff
@@ -37,6 +38,35 @@ func Handle(req []byte) string {
...
+	if readBool("validate_customers") {
+		customersURL := os.Getenv("customers_url")
+
+		customers, getErr := getCustomers(customersURL)
+		if getErr != nil {
+			return getErr.Error()
+		}
+
+		customer := sdk.Customer{}
+		json.Unmarshal(req, &customer)
```
bartsmykla: There is missing error handling

github-event/handler.go
```diff
@@ -37,6 +38,35 @@ func Handle(req []byte) string {
...
+
+		found = validCustomer(customers, customer.Sender.Login)
+
+		if !found {
```
bartsmykla: Instead of `if !found {` you can do:
```
if !validCustomer(customers, customer.Sender.Login) {
```

github-event/handler.go
```diff
@@ -238,3 +268,56 @@ func forward(req []byte, function string, headers map[string]string) (string, in
...
+	for _, customer := range customers {
+		if len(customer) > 0 &&
+			strings.EqualFold(customer, owner) {
+			found = true
+			break
+		}
+	}
+
+	return found
+}
```
bartsmykla: I would write it:
```
func validCustomer(customers []string, owner string) bool {
	for _, customer := range customers {
		if len(customer) > 0 && strings.EqualFold(customer, owner) {
			return true
		}
	}

 	return false
}
```

github-event/handler.go
```diff
@@ -238,3 +268,56 @@ func forward(req []byte, function string, headers map[string]string) (string, in
...
+
+	if res.Body != nil {
+		defer res.Body.Close()
+
+		pageBody, _ := ioutil.ReadAll(res.Body)
+		customers = strings.Split(string(pageBody), "\n")
+
+		for i, c := range customers {
+			customers[i] = strings.ToLower(c)
+			customers[i] = strings.TrimSuffix(c, "\r")
```
bartsmykla: There is something wrong, you are assigning value and then immediately overriding it

github-event/handler.go
martindekov: Done ^

github-event/handler.go
martindekov: Yep this was irrational 😄

github-event/handler.go
martindekov: Good idea!

martindekov:
I amended some fixes, I haven't looked into the function I just moved the logic out. Also added more detailed reports since those were only from the requests. Thanks for the feedback @bartsmykla really helped out

PR #109, opened 2018-08-23
Add support for gitlab push and system hooks

martindekov:
Adding gitlab-push function which can be added to webhook route
to handle push or system hooks from gitlab

Signed-off-by: Martin Dekov (VMware) <mdekov@vmware.com>

## Description

Handling push-events which deploy your function from repo in git-lab.

## How Has This Been Tested?

1. Deploy the function in your OF-cloud instance `faas-cli deploy -f gitlab.yml`
2. Configure your gitlab webhook route to `<gateway_address>/functions/gitlab-push`
3. Make function somewhere `faas-cli new <funcname> --lang=go --prefix=<your dockerhub name>`
4. ```mv <funcname>.yml stack.yml```
5. Push function to your gitlab repo
6. Check function deployed in `<gateway>/functions/overview?username=<name of account in gitlab>`

## Checklist:

I have:

- [ ] updated the documentation if required
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [ ] added unit tests

PR #274, opened 2018-10-08
[WIP] Add gitlab-event

martindekov:
## Description

We need to handle function installation/deletion on different scenarios which we currently don't with gitlab this extends #109 .

- [x] Garbage collect on tag removal

- [x] Installation on present tag

- [ ] Install multiple functions by labeling instance?

- [ ] Handle non-push events to install/uninstall functions

## How Has This Been Tested?

On my own instance

They will be able to use gitlab

## Checklist:

I have:

- [ ] added unit tests


martindekov:
Messed up commits.

Issue #93, opened 2018-08-09
git-tar deploy does not return error on non-success http status code


burtonr:
When looking through the logs in `git-tar`, I found that my function deployment returned a 500 error, however, there is no error checking for the deploy status

```
git-tar.1.8qapnsejzztn@playground    | Deploying service - spacex-info
git-tar.1.8qapnsejzztn@playground    | 2018/08/09 02:11:25 v1.53cf333d41169248e81195a8811570348ca18329
git-tar.1.8qapnsejzztn@playground    | Service deployed  spacex-info 500 Internal Server Error burtonr
git-tar.1.8qapnsejzztn@playground    | 2018/08/09 02:11:57 Functions owned by burtonr:
git-tar.1.8qapnsejzztn@playground    |  []Garbage collection ran for burtonr/test-function - 0 functions deleted.
git-tar.1.8qapnsejzztn@playground    | 
git-tar.1.8qapnsejzztn@playground    | 2018/08/09 02:11:58 Duration: 33.223507 seconds
```

The `c.Do(httpReq)` does not throw an error, so this goes unnoticed as the request was successful, but the response status is not success.

See here: https://github.com/openfaas/openfaas-cloud/blob/master/git-tar/function/ops.go#L252

ivanayov:
Just a notice if someone takes this: for both Do and Get, a non-2xx response doesn't cause an error. So this needs an additional check from our side.

martindekov:
Derek close: We have this covered up now https://github.com/openfaas/openfaas-cloud/blob/ab3cb6fef86cf4da7ea2ff69cbc331af553695f5/git-tar/function/ops.go#L409

PR #279, opened 2018-10-11
Edit git-tar to deploy functions in parallel

### LINKED ISSUE #252: git-tar should invoke deploy (buildshiprun) in parallel
## Expected Behaviour

git-tar should invoke deploy (buildshiprun) in parallel

## Current Behaviour

If a stack has 10 functions they are built in serial one at a time. Use an errgroup (see of-builder for an example of this) to build them parallel, but with a limit applied so that the concurrency is manageable - i.e. 2 or 4 at once.

The errgroup can be used to synchronise and collect any errors that come up.

Serial deployment loop:

https://github.com/openfaas/openfaas-cloud/blob/master/git-tar/function/ops.go#L319

errgroup usage:

https://github.com/openfaas/openfaas-cloud/blob/master/of-builder/main.go#L196

ivanayov:
With this change git-tar will POST to buildshiprun in parallel,
with a restriction on maximum 4 functions to be processed at once
(If required, this can be made configurable)

It will wait for all functions to be deployed, but will return the first error only (this is how errgroup behaves). The reason to implement it this way is that it will behave as a typical build system. Second if someone has multiple functions in a single repo, this means that they should be dependent on each other and reporting the first error and fixing the functions one by one in this context makes sense.
The other option is to keep back `failedFunctions` and synchronise it, but it's aways better not to add synchronised array when it's possible, given this opens a field for future bugs.

Closes #252 , #93

## How Has This Been Tested?

Tested on this repo - [ivanayov/ofc-pr-testing/runs/26567991](https://github.com/ivanayov/ofc-pr-testing/runs/26567991)
The functions were building in parallel and the status was updated successfully at the end.
With one function failing, it waits for all and reports failing status for that one.
With more than one failing - the same, but reports one of them as failing.

For repo with 1 function, works as normal, but starts the function build few seconds later.

For repo with 0 functions fails with `unable to find stack.yml`

Not affected

## Checklist:

I have:

- [ ] added unit tests - not compatible for the use case, needs live testing instead

### DIFF
--- a/git-tar/function/ops.go
+++ b/git-tar/function/ops.go
@@ -3,6 +3,7 @@ package function
 import (
 	"archive/tar"
 	"bytes"
+	"context"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -20,12 +21,15 @@ import (
 
 	"code.cloudfoundry.org/bytefmt"
 	"github.com/openfaas/faas-cli/schema"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/alexellis/derek/auth"
 	"github.com/alexellis/hmac"
 	"github.com/openfaas/faas-cli/stack"
 	"github.com/openfaas/openfaas-cloud/sdk"
 )
+
+const deployInParallelLimit = 2
 
 type tarEntry struct {
 	fileName     string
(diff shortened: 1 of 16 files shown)


git-tar/function/ops.go
```diff
@@ -316,91 +317,98 @@ func deploy(tars []tarEntry, pushEvent sdk.PushEvent, stack *stack.Services, sta
 	c := http.Client{}
 	gatewayURL := os.Getenv("gateway_url")
 
+	sem := make(chan int, 4)
```
alexellis: sem?

git-tar/function/ops.go
```diff
@@ -316,91 +317,98 @@ func deploy(tars []tarEntry, pushEvent sdk.PushEvent, stack *stack.Services, sta
...
+	eg, _ := errgroup.WithContext(context.Background())
+
 	for _, tarEntry := range tars {
-		fmt.Println("Deploying service - " + tarEntry.functionName)
+		sem <- 1
+		eg.Go(func() error {
```
alexellis: This is too nested now, I'd like to see it broken down and broken out.

git-tar/function/ops.go
ivanayov: This is for semaphore. Does it look clear or I should better rename?

alexellis: It's not a semaphore it's a channel. Consider a more meaningful name. workChannel? (I don't know)

git-tar/function/ops.go
ivanayov: It's addressed on next lines

git-tar/function/ops.go
```diff
@@ -306,103 +308,122 @@ func clone(pushEvent sdk.PushEvent) (string, error) {
...
-	failedFunctions := []string{}
-	owner := pushEvent.Repository.Owner.Login
-	repoName := pushEvent.Repository.Name
-	url := pushEvent.Repository.CloneURL
-	afterCommitID := pushEvent.AfterCommitID
-	installationID := pushEvent.Installation.ID
-
-	c := http.Client{}
-	gatewayURL := os.Getenv("gateway_url")
+	sem := make(chan int, 4)
```
alexellis: Let's get this variable name changed please @ivanayov

ivanayov:
@alexellis I've addressed your comment.
The PR now is fixed

alexellis:
Needs rebasing and another review.

git-tar/function/ops.go
```diff
@@ -306,109 +308,128 @@ func clone(pushEvent sdk.PushEvent) (string, error) {
...
-	url := pushEvent.Repository.CloneURL
-	afterCommitID := pushEvent.AfterCommitID
-	installationID := pushEvent.Installation.ID
-	sourceManagement := pushEvent.SCM
-	privateRepo := pushEvent.Repository.Private
-	repositoryURL := pushEvent.Repository.RepositoryURL
-
-	c := http.Client{}
-	gatewayURL := os.Getenv("gateway_url")
+	workCh := make(chan int, 4)
```
alexellis: `4` should be a const and also let's set it to `2` so we don't cripple the system.

ivanayov:
@rosti thanks for the code review. Can you please add a comment?

rosti:
The current version looks OK from concurrency point of view.

The only obvious things are a bit stylistic. I see that you use the `x, _ := ...` construct very often. You can reduce that to just `x := ...`.

ivanayov:
@rosti this syntax does not work for function call

ivanayov:
There's an issue with GitHub I think. Travis build is [passing](https://travis-ci.org/openfaas/openfaas-cloud/builds/451913906#L1654), but the status stays as pending for ~24h now.

burtonr:
@ivanayov I re-ran the build in Travis, and the check was updated successfully!

Just need to ping @alexellis to make sure his changes were addressed so this can be merged

alexellis:
Bumping up the issue as @LucasRoesler mentioned this recently. 

@ivanayov please do an e2e test in a live cluster using repos with 0, 1 and multiple functions.

alexellis:
Derek add label: blocked

alexellis:
Blocked awaiting more testing.

ivanayov:
I'm on it

alexellis:
Given the potential issues found, let's park this for the time being.

alexellis:
Derek close: due to issues found during testing.

Issue #210, opened 2018-09-22
Assign "SCM" field to functions via github-push


alexellis:
## Expected Behaviour

We should assign an "SCM" field to functions via `github-push` so that we can use this information for re-constituting links later on.

For example:

```go
prefix + "scm" = "github"
```

## Current Behaviour

We assume only GitHub, but have support for GitLab via #101 - this means dashboard links etc are wrong when using GitLab - they are instead directed to GitHub.

For GitLab we may need to have a further field such as "SCM address" since I expect the integrated GitLab instance to be something running at an address like: https://gitlab.team-serverless.xyz. GitHub will always be https://github.com/

## Possible Solution

Please provide some ideas / suggestions on above.

alexellis:
cc @bartsmykla @martindekov

alexellis:
@martindekov please can you take this?

martindekov:
Yeah

martindekov:
Derek assign: me

alexellis:
Source Control Management i.e. GitHub / GitLab / BitBucket

martindekov:
I was wondering if it is better approach to just handle whole URL to the repository path instead of constructing the URL? GitLab provides the repo URL in its payload:
```
...
      "web_url":"https://gitlab.team-serverless.xyz/martindekov/gitlab-tag-event",
...
```
and github provides the URL in its payload:
```
...
    "html_url": "https://github.com/martindekov/gocloudfunc",
...
```
Of course this will require change in the dashboards but they would be minor and @bartsmykla agreed on helping me integrating this into the dashboard.

PR #281, opened 2018-10-11
Edit findStackFile to form raw URL depending on GitHub/GitLab

### LINKED ISSUE #276: Edit git-tar raw URL stack file optimisation to support both GitHub/GitLab
This commit https://github.com/openfaas/openfaas-cloud/commit/aa5828702f0e82f5b1070cb1c1b73ff908ceb4e3 adds optimization for git-tar but I think the format of the URL in this line https://github.com/openfaas/openfaas-cloud/commit/aa5828702f0e82f5b1070cb1c1b73ff908ceb4e3#diff-6972f28fc33acdd0e2082babf8ec8405R276 should be moved away from git-tar possibly sent through github-push or github-event to the git-tar so the function itself is non dependent on github or any SCM

## Expected Behaviour

URL to be formatted in github-push or github-event rather than git-tar. Move away SCM specific field out of git-tar.

## Current Behaviour

Git-tar suggest only github raw formatting

## Possible Solution

Format the URL in github-event or github-push function and send it to git-tar.

It is not a bug per se

## Context

It is currently blocking #275 since the format of the URL is wrong for gitlab. And the workaround I am currently using is to set the `private_repo:true` all the time for gitlab, which is not the case. Also I think it is a good thing to keep functions down from git-tar away from SCM specific code.

ivanayov:
The raw URL for stack.yml was previously supporting only GitHub, so the build was failing for GitLab. With this change raw URL is properly formed depending on the git hosting solution  (GitHub/GitLab)

Closes #276

## How Has This Been Tested?

Tested by @martindekov on GitLab
I tested on GitHub:
1. Succeeds when stack.yml is available
2. Fails when not with `unable to find stack.yml`
3. After reproducing error with empty raw url (by using an image for `github-push` without the changes) fails with `given stack file raw URL is empty`. This resolves previous problem in behaviour - if fileSTackFile fails with error, there were no status reported.

Not impacted

## Checklist:

I have:

- [ ] added unit tests N/A

### DIFF
--- a/git-tar/function/handler.go
+++ b/git-tar/function/handler.go
@@ -54,10 +54,15 @@ func Handle(req []byte) []byte {
 	statusEvent := sdk.BuildEventFromPushEvent(pushEvent)
 	status := sdk.BuildStatus(statusEvent, sdk.EmptyAuthToken)
 
-	hasStackFile, getStackFileErr := findStackFile(&pushEvent.Repository)
+	hasStackFile, getStackFileErr := findStackFile(&pushEvent)
 
 	if getStackFileErr != nil {
-		log.Printf("cannot fetch stack file %s", getStackFileErr.Error())
+		msg := fmt.Sprintf("cannot fetch stack file %s", getStackFileErr.Error())
+
+		status.AddStatus(sdk.StatusFailure, msg, sdk.StackContext)
+		reportStatus(status)
+
+		log.Printf(msg)
 		os.Exit(-1)
 	}
 
@@ -266,14 +271,18 @@ func reportStatus(status *sdk.Status) {
 // available via the CDN. Note: given that the CDN has a 5-minute timeout - this optimization
 // may have the undesired effect of preventing a user from deploying within a 5 minute window
 // of renaming an incorrect "function.yml" to "stack.yml"
-func findStackFile(repo *sdk.PushEventRepository) (bool, error) {
+func findStackFile(pushEvent *sdk.PushEvent) (bool, error) {
 
 	// If using a private repo the file will not be available via the git-raw CDN
-	if repo.Private {
+	if pushEvent.Repository.Private {
 		return true, nil
 	}
 
-	addr := fmt.Sprintf("https://raw.githubusercontent.com/%s/%s/master/stack.yml", repo.Owner.Login, repo.Name)
+	addr, err := getRawURL(pushEvent.SCM, pushEvent.Repository.RepositoryURL, pushEvent.Repository.Owner.Login, pushEvent.Repository.Name)
+	if err != nil {
+		return false, err
+	}
+
 	req, _ := http.NewRequest(http.MethodHead, addr, nil)
 	log.Printf("Stack file request: %s", addr)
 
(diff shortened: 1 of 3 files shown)


martindekov:
Tested by deploying image provided by @ivanayov  with the change the following scenario deployed gitlab-push in `gitlab.yml` with option: 
```
      private_repo: false
```
(before had to workaround this by setting `private_repo: true`) and it built my function successfully.
Also amended #109 to use the solution.

alexellis:
For GitLab could we look at the "Repo URL" from the push event and use that to create the URL for the stack.yml?

martindekov:
This is connected with Issue that I am assigned to #210 and I will open PR and explain here how to extend/remake this PR with the change.

martindekov:
So @ivanayov I have opened PR #291 which fixes the issue pointed above in the proper way.
In `git-tar` get `event.SCM` field and check `github-push`? if so construct in the URL in github manner. For gitlab since we have custom URL I might need to add another field/function to construct the URL I think https://github.com/martindekov/openfaas-cloud/commit/20c6b09edf03a329aff676195f1827ed1a7949f2 may be used to construct the URL

git-tar/function/handler.go
```diff
@@ -266,14 +271,26 @@ func reportStatus(status *sdk.Status) {
 // available via the CDN. Note: given that the CDN has a 5-minute timeout - this optimization
 // may have the undesired effect of preventing a user from deploying within a 5 minute window
 // of renaming an incorrect "function.yml" to "stack.yml"
-func findStackFile(repo *sdk.PushEventRepository) (bool, error) {
+func findStackFile(pushEvent sdk.PushEvent) (bool, error) {
```
alexellis: Why did we remove the pointer reference? Should add it back to not copy by value.

alexellis: How has this been unit tested? I'd like to see unit tests for this as part of the PR.

git-tar/function/handler.go
```diff
@@ -296,3 +305,22 @@ func findStackFile(repo *sdk.PushEventRepository) (bool, error) {
 
 	return false, nil
 }
+
+func getRawURL(scm string, repositoryURL string, repositoryOwnerLogin string, repositoryName string) (string, error) {
+
+	addrPrefix := ""
+	rawPath := ""
+	switch scm {
```
alexellis: `scm` stands for source control management - so this should be only `github` or only `gitlab`. Not sure why we are adding the text `-push` here? Please change.

git-tar/function/handler_test.go
```diff
@@ -0,0 +1,53 @@
+package function
+
+import (
+	"testing"
+)
+
+func Test_getRawURL(t *testing.T) {
```
alexellis: Good to see the tests.

git-tar/function/handler.go
ivanayov: These are the values that come from pushEvent.SCM. 
We need to change them in order to omit the `-push` suffix here

martindekov: I added the change but directly with `Source` constant. This can be quickly changed.

ivanayov:
@alexellis I've updated the scm field

git-tar/vendor/github.com/openfaas/openfaas-cloud/sdk/events.go
```diff
@@ -20,6 +20,7 @@ type PushEvent struct {
 	Repository    PushEventRepository
 	AfterCommitID string `json:"after"`
 	Installation  PushEventInstallation
+	RepositoryURL string `json:"url"`
```
martindekov: This should be inside `PushEventRepository` struct

martindekov: This should be inside `PushEventRepository` struct

ivanayov:
@martindekov , I've updated the url

alexellis:
Merged

Issue #283, opened 2018-10-12
UI: Dashboard - create logout endpoint


bartsmykla:
Dashboard doesn't have access to cookie `.system.domain.com` so we can't remove it just from UI. I think that the good idea will be to create simple endpoint from `auth.system.domain.com/logout` which would be used when we click `logout` button. Related to https://github.com/openfaas/openfaas-cloud/issues/256

## Expected Behaviour
Expose `/logout` endpoint which will "remove" cookie.

## Current Behaviour
There is no way to remove cookie from the dashboard

## Possible Solution
Expose `/logout` endpoint which will "remove" cookie.

bartsmykla:
Derek assign: me

PR #284, opened 2018-10-12
Expose /logout endpoint

bartsmykla:

## Description
Closes: https://github.com/openfaas/openfaas-cloud/issues/283

Exposed /logout endpoint will remove openfaas_cloud_token cookie

## How Has This Been Tested?

Manually used endpoint to remove cookie few times.

They are not.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/auth/handlers/logout.go
+++ b/auth/handlers/logout.go
@@ -1,0 +1,22 @@
+package handlers
+
+import (
+	"net/http"
+	"time"
+)
+
+// MakeLoginHandler creates a handler for logging out
+func MakeLogoutHandler(config *Config) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		http.SetCookie(w, &http.Cookie{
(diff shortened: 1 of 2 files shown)


alexellis:
Can this be be handled by the dashboard instead? The button will be in the dashboard and the dashboard should not know about the auth service.

bartsmykla:
With current cookie it's not possible, because it's for domain `.system.domain.com`(so for `auth.system.domain.com` it's perfect) abutd dashboard is accessible at `system.domain.com/dashboard`

bartsmykla:
Hmmm. But I think now, that we can set the cookie domain for `system.domain.com`, let me see if it'll work.

bartsmykla:
@alexellis actually I was wrong. I didn't see that we are setting up `HttpOnly` flag for a cookie, so I need to set it to false to be able to remove cookie from dashboard.

bartsmykla:
Derek close

PR #286, opened 2018-10-12
Add struct to sdk that handles GitLab payload

martindekov:
Adding struct that handles GitLab payload cause currently
PR #109 that handles push events from GitLab use copied
vendor folder which is not the right way to vendor code
the PR above need proper vendoring

## Description

Currently PR #109 "vendors" GitLab sdk by copying `vendor` folder from other project and add the struct manually and I would like to vendor the code the proper way so when this PR is merged and new release of SDK is made I can make code in the above PR vendored properly

## How Has This Been Tested?

N.A. Exist in PR #109 and used multiple times for testing but in the PR the change is "vendored" by copying the vendor folder from other projects.

N.A. this is part of bigger change

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/sdk/events.go
+++ b/sdk/events.go
@@ -41,6 +41,28 @@ type Event struct {
 	Private        bool              `json:"private"`
 }
 
+// PushEvent as received from GitLab
+
+type GitLabPushEvent struct {
+	Ref              string           `json:"ref"`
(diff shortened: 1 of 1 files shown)


alexellis:
Merged on Sofia community call.

PR #309, opened 2018-10-24
Add gitlab-status function

martindekov:
## Description

This change adds support when reporting the stage of the function through status for GitLab

The concept:

gitlab status

## How Has This Been Tested?

Updated the change on image `martindekov/of-git-tar:gitlab-status` and `martindekov/of-buildshiprun:gitlab-status` to enable reporting for those functions.

And the `gitlab-status` function itself is configured in `gitlab.yml` file the following way:

```
provider:
  name: faas
  gateway: http://127.0.0.1:8080

functions:
  gitlab-status:
    lang: go
    handler: ./gitlab-status
    image: martindekov/gitlab-status:latest
    environment:
      write_debug: true
    environment_file:
      - gateway_config.yml
    secrets:
      - gitlab-api-token
      - payload-secret
```
Where `payload-secret` is the already existing internal secret `gitlab-api-token` is GitLab Access Token that is generated in your GitLab Account and gives access to the API.

Will be able to report the function building status to GitLab

## Checklist:

I have:

- [x] added unit tests


buildshiprun/handler.go
```diff
@@ -94,7 +94,14 @@ func Handle(req []byte) string {
 		sdk.PostAudit(auditEvent)
 
 		status.AddStatus(sdk.StatusFailure, err.Error(), sdk.BuildFunctionContext(event.Service))
-		reportStatus(status)
+		if event.SCM == "github" {
+			reportStatus(status)
```
ivanayov: Should better return error from both methods and log here, or log within the methods on both places.

Why not move the SCM check within a single reportStatus method?

buildshiprun/handler.go
```diff
@@ -115,7 +122,14 @@ func Handle(req []byte) string {
...
```
ivanayov: Same here?

buildshiprun/handler.go
martindekov: This is not function I wrote and it was like that before so I left it

martindekov: I will change the name `reportStatus` to `reportGitHubStatus` and move the logs of `reporGitLabStatus` inside the log of the function body. Is that ok @ivanayov

martindekov:
@ivanayov referenced you comments and amended fix which refactors the code to be more clear while reporting statuses.

gitlab-status/handler.go
```diff
@@ -0,0 +1,132 @@
...
+
+func gitLabURLBuilder(status *sdk.Status) string {
+	delimeterPlace := 3
+	delimeterValue := "/"
+	baseURL := getURLbyDelimeter(status.EventInfo.URL, delimeterPlace, delimeterValue)
+	routeURL := fmt.Sprintf("/api/v4/projects/%d/statuses/%s?", status.EventInfo.InstallationID, status.EventInfo.SHA)
+	newURL := append(strings.Split(baseURL, ""), routeURL)
+	wholeURL := strings.Join(newURL, "")
+	return wholeURL
+}
```
bartsmykla: It should be unit tested

gitlab-status/handler.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		if symbol == delimeter {
+			slash++
+		}
+	}
+
+	sliceURL = append(sliceURL[:last-1], sliceURL[cap(sliceURL):]...)
+	baseURL = strings.Join(sliceURL, "")
+	return baseURL
+}
```
bartsmykla: I think that it should be done using `http/url`

gitlab-status/handler.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if state == "failure" {
+		state = "failed"
+	}
+
+	parameters := url.Values{}
+	parameters.Add("state", state)
+	parameters.Add("description", desc)
+	parameters.Add("context", context)
+	theURL.RawQuery = parameters.Encode()
```
bartsmykla: It should be unit tested, add case with URL (as a first parameter) which contains query parameters .

gitlab-status/handler.go
```diff
@@ -0,0 +1,132 @@
...
+	if reqErr != nil {
+		return reqErr
+	}
+	req.Header.Set("PRIVATE-TOKEN", token)
+	client := &http.Client{}
+	resp, clientErr := client.Do(req)
+	if clientErr != nil {
+		return clientErr
+	}
+	resp.Body.Close()
```
bartsmykla: I'm not sure if you should check first if body is nil or not. This comment is for me to check it.

gitlab-status/handler.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf("couldn't get payload-secret: %t", err)
+	}
+
+	xCloudSignature := os.Getenv("Http_X_Cloud_Signature")
+
+	err = hmac.Validate(req, xCloudSignature, payloadSecret)
+
+	if err != nil {
+		return err
+	}
```
bartsmykla: It should be added to sdk, but @martindekov is saying that this function appears everywhere. I would write it also:
```
if err := hmac.Validate(req, xCloudSignature, payloadSecret); err != nil {
 	return err
}
```

gitlab-status/handler_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	for _, test := range stringStruct {
+		t.Run(test.Title, func(t *testing.T) {
+			leftoverString := getURLbyDelimeter(test.testString, test.position, test.delimeter)
+			if leftoverString != test.expectedString {
+				t.Errorf("Expected string: `%s` got: `%s`", test.expectedString, leftoverString)
+			}
+		})
+	}
+}
```
bartsmykla: You should add test cases what will happen when there is no delimiters or there are only delimiters in string

buildshiprun/handler.go
```diff
@@ -422,7 +422,19 @@ func enableStatusReporting() bool {
 	return os.Getenv("report_status") == "true"
 }
 
-func reportStatus(status *sdk.Status) {
+func reportStatus(status *sdk.Status, SCM string) {
+	if SCM == "github" {
```
bartsmykla: I think that we should check if SCM is github or gitlab, and if not, we should return error about not supporting that SCM

git-tar/function/ops.go
```diff
@@ -493,3 +493,75 @@ func getShortSHA(sha string) string {
...
+func reportStatus(status *sdk.Status, SCM string) {
+	if SCM == "github" {
+		reportGitHubStatus(status)
+	} else {
+		reportGitLabStatus(status)
+		for i := range status.CommitStatuses {
+			delete(status.CommitStatuses, i)
+		}
+	}
+}
```
bartsmykla: the same suggestion as in buildshiprun comment

alexellis:
I can see a conflict on: git-tar/function/ops.go

git-tar/function/ops.go
```diff
@@ -493,3 +493,75 @@ func getShortSHA(sha string) string {
...
 	return sha[:7]
 }
+
+func reportStatus(status *sdk.Status, SCM string) {
+	if SCM == "github" {
+		reportGitHubStatus(status)
+	} else {
+		reportGitLabStatus(status)
+		for i := range status.CommitStatuses {
+			delete(status.CommitStatuses, i)
```
alexellis: I think we can avoid the deletions by creating a new status object for the final stack status. Can you try that out?

ivanayov:
Needs rebasing as well

gitlab-status/handler_test.go
martindekov: They will be overwritten

martindekov: But I removed the use of that function

gitlab-status/handler.go
martindekov: It will overwrite them. I  can open an issue since we really don't use the function in any other way than status reporting. I can't think (for now) where we will use/send URL to the function with parameters

gitlab-status/handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+			url:         "https://some.random.url/api/v4/projects/3/statuses/99a7c6009c43cca39c61977ab8d3abdd13d7b111",
+			state:       "stack-deploy",
+			desc:        "some description",
+			context:     "context",
+			expectedURL: "https://some.random.url/api/v4/projects/3/statuses/99a7c6009c43cca39c61977ab8d3abdd13d7b111?context=context&description=some+description&state=stack-deploy",
+		},
+		{
+			title:       "Everything is set right and state is failure",
+			url:         "https://some.random.url/api/v4/projects/3/statuses/99a7c6009c43cca39c61977ab8d3abdd13d7b111",
+			state:       "failure",
```
ivanayov: What will happen if one or all of these is empty? Let's add more test cases.

gitlab-status/handler.go
ivanayov: Can be addressed in another PR

git-tar/function/ops.go
ivanayov: Breaking change. Needs careful testing before update.

gitlab-status/handler.go
martindekov: Yes you are right we should check for nil(actually just got error from that kind)

gitlab-status/handler.go
martindekov: Removed the function altogether and remade it added unit test for it

gitlab-status/handler.go
martindekov: Used the `http/url` as you suggested

gitlab-status/handler.go
martindekov: Amended fix for this

buildshiprun/handler.go
martindekov: Done

git-tar/function/ops.go
martindekov: Also done

git-tar/function/ops.go
martindekov: Amended fix as suggested ^

gitlab-status/handler_test.go
martindekov: Added that scenario ^

buildshiprun/handler.go
```diff
@@ -437,7 +455,19 @@ func enableStatusReporting() bool {
 	return os.Getenv("report_status") == "true"
 }
 
-func reportStatus(status *sdk.Status) {
+func reportStatus(status *sdk.Status, SCM string) error {
+	if SCM == "github" {
+		reportGitHubStatus(status)
+	} else if SCM == "gitlab" {
```
alexellis: Please use consts for `gitlab` etc

martindekov: Done

buildshiprun/handler.go
```diff
@@ -531,3 +566,48 @@ func getMemoryLimit() string {
...
+	}
+	if res.StatusCode != http.StatusOK {
+		log.Printf("unexpected status code: %d", res.StatusCode)
+	}
+
+	_, bodyErr := ioutil.ReadAll(res.Body)
+	if bodyErr != nil {
+		log.Printf("unexpected error while reading response body: %s", bodyErr.Error())
+	}
+	status.CommitStatuses = make(map[string]sdk.CommitStatus)
```
alexellis: Required: Let's do this with `status.Clear()` instead.

Optional: Consider using a pointer reference instead of the value, so that we don't by accident update a value and miss the update.

PR #264, opened 2018-10-01
Add support for custom templates

martindekov:
Enable users which have their own instance of OF-cloud
to add their own template repositories through configuration
rather than manually enter in the code and applying changes
to the right place

Signed-off-by: Martin Dekov (VMware) <mdekov@vmware.com>

## Description

This Closes #229. I went with the approach that separates the repositories with comma
```
custom_templates: "https://github.com/single/template.git, https://github.com/another/template.git"
```
though if requested can find a way to change this maybe with 
```
custom_templates:
  - https://github.com/single/template.git
  - https://github.com/another/template.git
```
But I can't think of a way to get those values from single key without having to parse yaml, which means I must copy the file inside the container and then parse ..

Went with separate function since the `fetchTemplates(something)` with that change would now format the templates AND fetch them and the function becomes too long for the thing it does(fetching) also inconvenient for testing.

## How Has This Been Tested?

Added tests also tested on my cloud with image `martindekov/of-git-tar:261` and modified configuration in `gateway_config.yml` with `custom_templates` variable like so 
```
...
  custom_templates: "https://github.com/openfaas-incubator/python-flask-template.git, https://github.com/openfaas-incubator/ruby-http.git"
...
```
the content of the folder inside the container is
```
$ ls template
csharp                   golang-http              java8                    node8-express            python-armhf             python3-flask
dockerfile               golang-http-armhf        node                     node8-express-armhf      python27-flask           python3-flask-armhf
go                       golang-middleware        node-arm64               php7                     python3                  ruby
go-armhf                 golang-middleware-armhf  node-armhf               python                   python3-armhf            ruby-http
```
Commented out the custom_templates like in `gateway_config.yml` like so
```
...
  #custom_templates: "https://github.com/openfaas-incubator/python-flask-template.git, https://github.com/openfaas-incubator/ruby-http.git"
...
```
and inside the container the templates folder contains
```
$ ls template
csharp                   golang-http              java8                    node8-express            python-armhf
dockerfile               golang-http-armhf        node                     node8-express-armhf      python3
go                       golang-middleware        node-arm64               php7                     python3-armhf
go-armhf                 golang-middleware-armhf  node-armhf               python                   ruby
```
## How are existing users impacted? What migration steps/scripts do we need?

N.A. This change is enhancement.

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
[...]

Issue #319, opened 2018-10-30
Validate custom template URLs


alexellis:
## Expected Behaviour

We should validate that custom template URLs are well-formed URLs.

http and https should both be valid.

## Current Behaviour

We don't

## Possible Solution

This is part of the review feedback of : https://github.com/openfaas/openfaas-cloud/pull/264

alexellis:
Please update the unit tests to handle errors when the URLs are invalid and reduce the level of nesting/indent as per - https://github.com/openfaas/openfaas-cloud/pull/320/files#r229253436

alexellis:
Additional comments: https://github.com/openfaas/openfaas-cloud/pull/320#discussion_r229260827

ivanayov:
@alexellis addressed your comments in #321

PR #320, opened 2018-10-30
 Follow up #264 (Add support for custom templates) 

ivanayov:
This addresses pull #264 comments

### DIFF
--- a/git-tar/function/ops.go
+++ b/git-tar/function/ops.go
@@ -44,9 +44,7 @@ func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error
 }
 
 func fetchTemplates(filePath string) error {
-	templateRepos := []string{"https://github.com/openfaas/templates",
-		"https://github.com/openfaas-incubator/node8-express-template.git",
-		"https://github.com/openfaas-incubator/golang-http-template.git"}
+	templateRepos := formatTemplateRepos()
 
 	for _, repo := range templateRepos {
 		pullCmd := exec.Command("faas-cli", "template", "pull", repo)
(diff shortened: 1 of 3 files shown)


git-tar/function/ops_test.go
```diff
@@ -100,3 +102,106 @@ func Test_getRepositoryURL_whenGetTokenReturnsError_WhenRepositoryIsPrivate(t *t
...
+		},
+		{
+			title:         "Templates with two added custom repositories without spaces",
+			envRepos:      "https://github.com/my-custom/repo.git,https://github.com/another/repo.git",
+			expectedRepos: append(formalRepos, ([]string{"https://github.com/my-custom/repo.git", "https://github.com/another/repo.git"})...),
+		},
+	}
+	for _, test := range tests {
+		os.Setenv("custom_templates", test.envRepos)
+		t.Run(test.title, func(t *testing.T) {
```
alexellis: This level of nesting is confusing. I think we can do better this will help with maintainability.

git-tar/function/ops.go
```diff
@@ -489,3 +487,24 @@ func importSecrets(pushEvent sdk.PushEvent, stack *stack.Services, clonePath str
 
 	return nil
 }
+
+// getShortSHA returns shorter version of git commit SHA
+func getShortSHA(sha string) string {
```
alexellis: Please remove this, it's now part of the SDK package.

git-tar/function/ops.go
```diff
@@ -489,3 +487,20 @@ func importSecrets(pushEvent sdk.PushEvent, stack *stack.Services, clonePath str
...
 	return nil
 }
+
+func formatTemplateRepos() []string {
+	templateRepos := []string{"https://github.com/openfaas/templates"}
+
+	if envTemplates := os.Getenv("custom_templates"); len(envTemplates) > 0 {
+		customTemplates := strings.Split(envTemplates, ",")
+		for _, repo := range customTemplates {
+			if strings.Contains(envTemplates, "https://") || strings.Contains(envTemplates, "http://") {
```
alexellis: This does not count as URL validation. You should be using https://godoc.org/net/url#URL.Parse from the Go stdlib.

I also can't see any error handling here. `func formatTemplateRepos() []string` should be: `func formatTemplateRepos() ([]string, error)`

Fix this up in the linked issue #319

Issue #171, opened 2018-09-13
UI: dashboard detail page - "run on my OpenFaaS" button


alexellis:
For the UI dashboard page I'd like to see a new button for users called "run on my OpenFaaS"

When you click on it a modal appears which you can copy/paste a stack.yml which is generated by the webpage using the Docker image name and function name.

It should also have a copy/paste command for the faas-cli to deploy.

Example:

Save as `stack.yml`:
```
provider:
  name: faas
  gateway: http://127.0.0.1:8080

functions:
  octo:
    skip_build: true
    image: ofcommunity/alexellis-go-fns-tester-octo:latest-665d959
```

Then run: 

```
faas-cli deploy
```

affix:
@alexellis It would be cool to have it have a ready to run one liner

like

```
$ curl https://url.tld/guid/stack.yml | faas-cli deploy
```

or allow the deploy command to take a URL to a stack

kenfdev:
Derek add label: area/dashboard

kenfdev:
How to integrate modals to the current dashboard is discussed in the following issue:

https://github.com/openfaas/openfaas-cloud/issues/172

ivanayov:
Derek assign: me

ivanayov:
Derek unassign: me

bartsmykla:
Derek assign: me

alexellis:
Derek add label: Hacktoberfest

alexellis:
Derek unassign: bartsmyka

alexellis:
Let's only assign work when actively working on it to prevent blocking participation from the community.

alexellis:
Derek unassign: bartsmykla

bartsmykla:
Derek assign: me

PR #326, opened 2018-11-01
Added Run on my OpenFaaS functionality

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/issues/171

## Description
- In function description page added button which clicked
opens modal with code which you need to run to quickly run
function in your OpenFaaS instance

## How Has This Been Tested?

Deployed on my own cluster end went to multiple different function pages checking if generated code is correct

This is just ui change, and doesn't need to any migration from user. It should just work  

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/dashboard/client/src/components/ModalRunOnMyOF/ModalRunOnMyOF.jsx
+++ b/dashboard/client/src/components/ModalRunOnMyOF/ModalRunOnMyOF.jsx
@@ -1,0 +1,118 @@
+import React, { Component } from 'react';
+import {
+  Modal,
+  ModalBody,
+  ModalHeader,
+  FormFeedback,
+  Form,
+  Input,
+  Button,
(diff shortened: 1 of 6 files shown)


dashboard/client/src/components/ModalRunOnMyOF/ModalRunOnMyOF.jsx
```diff
@@ -0,0 +1,118 @@
...
+    })
+  };
+
+  closeModal() {
+    this.setState({ active: false });
+  }
+
+  render() {
+    const { shortName, image, gitOwner } = this.props.fn;
+    const code = `{
```
alexellis: In a follow-up PR could you extract this into a new function? (perhaps add a unit test too?)

PR #307, opened 2018-10-24
Add gitlab-event function

martindekov:
## Description

This function is the front-facing function which get events from GitLab

Conceptual:

asdzxc

## How Has This Been Tested?

Have `gitlab.yml` file looking like 
```
provider:
  name: faas
  gateway: http://127.0.0.1:8080

functions:
  system-gitlab-event:
    lang: go
    handler: ./gitlab-event
    image: martindekov/gitlab-event:0.0.1
    labels:
      openfaas-cloud: "1"
      role: openfaas-system
    environment:
      content_type: text/plain
      validate_customers: true
      validate_token: true
      write_debug: true
      read_debug: true
      gitlab_instance: "https://gitlab.dedal.io/"
    environment_file:
      - gateway_config.yml
    secrets:
      - gitlab-webhook-secret
      - payload-secret
      - gitlab-api-token
```

Where secrets are:
 
* `gitlab-webhook-secret` is the secret provided when creating `System Hook`

* `payload-secret` is the internal trust secret

* `gitlab-api-token` is the api Access token to check if `openfaas-cloud` tag is present which means the project is part of the openfaas-cloud "installation"

And env vars are:

* `validate_customers: true` for customer of the instance validation

*  `validate_token: true` to match `gitlab-webhook-secret` with the secret sent with the `System Hook`

* `gitlab_instance: "https://gitlab.dedal.io/"` the base URL of your GitLab instance since the delete/update events don't send the URL

## How are existing users impacted? What migration steps/scripts do we need?

They will have GitLab front-facing function which triggers events

## Checklist:

I have:

- [x] updated the documentation and/or roadmap (if required)
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [x] added unit tests

PR #327, opened 2018-11-02
Add full GitLab Integration

### LINKED ISSUE #225: Feature: GitLab CI/CD support
Part 1 - After a planning meeting with the FaaS team, we want to restrict gitlab-push to only work for repos that are tagged as "openfaas-cloud". Builds upon #109 which is working 👍 

Part 2 - We need to garbage collect repos / functions when the tag is removed.
Tag events are sent via System Hooks.

Part 3 - we will need to document how to deploy OpenFaaS Cloud for GitLab as an alternative to GitHub - this needs to be structured appropriately in the docs/README.md file.

### LINKED ISSUE #249: Tracking Feature: for GitLab
This issue tracks the feature for GitLab support.

Priority:

* #225 Feature: GitLab CI/CD support @martindekov 
* #224 Feature: Dashboard/API: support GitLab or other SCM priority/high @martindekov @bartsmykla 
* #223 Feature: OAuth 2.0 login via GitLab @bartsmykla / @alexellis 
* TBC Feature: GitLab CI/CD for private repos @martindekov

martindekov:
## Description

This pull request adds PR #308 and #307 in one against the master branch

Closes #225 
Closes #249 

## How Has This Been Tested?

Tested in #308 and #307 

They will be able to have GitLab on prem.

## Checklist:

I have:

- [x] added unit tests


gitlab-push/handler.go
```diff
@@ -0,0 +1,221 @@
...
+
+const (
+	PrivateRepo  = 00
+	InternalRepo = 10
+	PublicRepo   = 20
+	Source       = "gitlab-push"
+	SCM          = "gitlab"
+)
+
+var audit sdk.Audit
```
alexellis: I'm confused why we have `audit` here and not passed to where it is needed?

martindekov: It is how it's done in `github-push` function and I followed the pattern. I can move it inside if necessary

docs/GITLAB.md
```diff
@@ -0,0 +1,124 @@
...
+echo -n "<your_token_here>" | docker secret create gitlab-webhook-secret -
+```
+
+
+The supported events are currently `push` and `project_update`/`project_destroy` through the System Hook so check the `Push events` event only and then `Add system hook`
+
+### Configure your Access Token
+
+This token is mandatory as it gives access to the API from which we take the tag and recognize if you have OpenFaaS Cloud installed, also provides us with a way to clone private/internal repositories and check the groups in which you participate.
+
```
alexellis: Suggest which user should do this - perhaps a service account or administrator?

martindekov: ok

git-tar/function/ops.go
```diff
@@ -274,16 +274,39 @@ func clone(pushEvent sdk.PushEvent) (string, error) {
 		return "", fmt.Errorf("cannot create user-dir: %s", userDir)
 	}
 
-	at := &githubAuthToken{
-		appID:          os.Getenv("github_app_id"),
-		installationID: pushEvent.Installation.ID,
-		privateKeyPath: sdk.GetPrivateKeyPath(),
+	var cloneURL string
+
+	if pushEvent.SCM == "github" {
```
LucasRoesler: This would be a good place to use a switch.

As it exapands it probably also makes sense to move the logic for each case to a separate method

gitlab-event/handler.go
```diff
@@ -0,0 +1,409 @@
...
+	"github.com/openfaas/openfaas-cloud/sdk"
+)
+
+// Source name for this function when auditing
+const (
+	Source              = "gitlab-event"
+	EventSource         = "System Hook"
+	PushEvent           = "push"
+	ProjectUpdateEvent  = "project_update"
+	ProjectDestroyEVent = "project_destroy"
```
LucasRoesler: Typo, should be `ProjectDestroyEvent`

git-tar/function/ops.go
martindekov: Yeah, this is a great Idea I will move it to separate functions

docs/GITLAB.md
martindekov: Done 👍

martindekov:
Thank you for the feedback @LucasRoesler @alexellis I added additional explanation for the Docs and moved the URL logic into its own functions

alexellis:
Merged. I will do a release of the new Docker images.

PR #329, opened 2018-11-05
Add redirect URL to the commit status (GitLab)

### LINKED ISSUE #323: GitLab commit status - add hyperlinks
## Expected Behaviour

The GitLab commit status can support a `target_url`, so we could link to the function endpoint in a successful status, or the dashboard / pipeline-logs in a failed status.

https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit

We could also add comments - for instance the failed container build logs for a bad function build. For instance your Golang function isn't valid due to formatting errors.

https://docs.gitlab.com/ee/api/commits.html#post-comment-to-commit

## Current Behaviour

The user should navigate to the dashboard manually.

martindekov:
Adding feature which will redirect the user to the
logs or function endpoint in case the function build
fails or succeeds

## Description

For GitLab commit statuses did not redirect to logs or endpoint, now they do in build failure/success

Closes #323

## How Has This Been Tested?

Performed the following manual tests:
Deployed three functions: working one, one with syntax error,
The functions were named:
* `swag` (working)
* `gitlab-playground` (syntax error)

Statuses were:
* `swag` (working) pointing to endpoint `https://martindekov.team-serverless.xyz/swag`
* `stack-deploy` always pointing to dashboard `https://system.team-serverless.xyz/dashboard/martindekov`
* `playground` (syntax error) pointing to logs `https://system.team-serverless.xyz/function/pipeline-log?repoPath=martindekov/gitlab-playground&commitSHA=551db6119c189f5e747f2c9ddc3732f2077dfef4`

I don't have the dashboard running so they can't be accessed but I follow the pattern

I have uploaded image with the change here: `martindekov/gitlab-status:3`

N.A.

## Checklist:

I have:

- [x] added unit tests


gitlab-status/handler.go
```diff
@@ -6,11 +6,16 @@ import (
...
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/alexellis/hmac"
 	"github.com/openfaas/openfaas-cloud/sdk"
 )
 
+const (
+	SystemSubDomain = "system"
```
alexellis: What is this? Perhaps it should not be hard-coded in a const but configured via environment like everything else?

gitlab-status/handler.go
```diff
@@ -32,8 +37,35 @@ func Handle(req []byte) string {
...
+		log.Printf("error while formatting endpoint URL: %s", endpointErr)
+	}
+
+	dashboardURL, dashboardErr := FormatDashboardURL(gatewayURL, &status.EventInfo)
+	if dashboardErr != nil {
+		log.Printf("the dashboard url failed: %s", dashboardErr.Error())
+	}
+
 	for _, commitStatus := range status.CommitStatuses {
-		reportErr := sendReport(url, token, commitStatus.Status, commitStatus.Description, commitStatus.Context)
```
alexellis: Can you please split the body of this for loop into a new function? sendCommitStatus or whatever makes sense.

gitlab-status/handler.go
```diff
@@ -112,3 +145,55 @@ func validateRequest(req []byte) (err error) {
 
 	return nil
 }
+
+func FormatLogsURL(gatewayURL string, event *sdk.Event) (string, error) {
```
alexellis: Curious why these methods are exported (starting with capital letters)?

gitlab-status/handler_test.go
```diff
@@ -73,47 +76,264 @@ func Test_appendParameters(t *testing.T) {
...
+			expectedURL: "https://system.o6s.io/dashboard/exampleusername/exampleservice/log?repoPath=exampleusername/examplerepository&commitSHA=98zxc823axcc",
+		},
+		{
+			title:       "gateway_public_url is set but the object is not populated",
+			event:       &sdk.Event{},
+			gatewayURL:  "https://cloud.o6s.io/",
+			expectedURL: "https://system.o6s.io/dashboard///log?repoPath=/&commitSHA=",
+		},
+		{
+			title: "gateway_public_url is not set right but the object is populated right",
```
alexellis: "not set right" - what does that mean when to a new contributor reading the code in 6 months? Can we improve the text?

gitlab-status/handler_test.go
```diff
@@ -73,47 +76,264 @@ func Test_appendParameters(t *testing.T) {
...
+			event: &sdk.Event{
+				Owner:      "exampleusername",
```
alexellis: What about just `user`? etc

gitlab-status/handler_test.go
```diff
@@ -73,47 +76,264 @@ func Test_appendParameters(t *testing.T) {
...
+		gatewayURL  string
+		expectedURL string
+	}{
+		{
+			title:       "gateway_public_url is set with slash",
+			gatewayURL:  "https://cloud.o6s.io/",
+			expectedURL: "https://system.o6s.io",
+		},
+		{
+			title:       "gateway_public_url is set without slash",
```
alexellis: without slash where? How about `trailing slash`? trailing = at end

gitlab-status/handler.go
martindekov: This is for the times the `gateway_public_url` is configured like `https://system.o6s.io` and we need `system` to replace the subdomain. I need to extend the functionality of this PR to support `pretty_url` environmental variable because I did not consider this. Totally forgot about this PR ..

gitlab-status/handler.go
martindekov: So move this block
```
		targetURL := endpointURL
		if commitStatus.Status != sdk.StatusSuccess {
			targetURL = logsURL
		}
		if commitStatus.Context == sdk.StackContext && commitStatus.Context != status.EventInfo.Service {
			targetURL = dashboardURL
		}
```
Into a new function?

gitlab-status/handler.go
martindekov: Though they would be needed in the future to be added to SDK but I will probably revert that

gitlab-status/handler_test.go
martindekov: Yeah you are right, I didn't think about this in that way, I will fix the titles

gitlab-status/handler_test.go
martindekov: Good suggestion ^

gitlab-status/handler_test.go
martindekov: Yep good proposal will add that

gitlab-status/handler_test.go
martindekov: Done, added better explanation

gitlab-status/handler_test.go
martindekov: Fixed that

gitlab-status/handler_test.go
martindekov: now I use `user`

gitlab-status/handler.go
martindekov: now the methods are not exported

gitlab-status/handler.go
martindekov: This logic ^ is now into its own method for convenience also added test

martindekov:
Tested with the changes manually and added unit tests to test the two functions that I added

alexellis:
One other part of my comment:

> Can you also mention all functions that have to be rebuilt as part of the impact?

Alex

gitlab-status/handler.go
```diff
@@ -112,3 +135,88 @@ func validateRequest(req []byte) (err error) {
...
+
+func getSubDomain(URL string) (string, error) {
+	parsedURL, parseErr := url.Parse(URL)
+	if parseErr != nil {
+		return "", fmt.Errorf("Unable to parse URL: %s", parseErr.Error())
+	}
+	subdomain := strings.Split(parsedURL.Host, ".")
+
+	//Host is www.world.org and subdomain would be www aka. 0th element of the slice
+	return subdomain[0], nil
```
alexellis: This needs range checking. What if there are 0 elements in the slice/array? then throw an error.

alexellis: Add a test-case for this too.

gitlab-status/handler.go
```diff
@@ -112,3 +135,88 @@ func validateRequest(req []byte) (err error) {
...
+	}
+	subDomain, err := getSubDomain(gatewayURL)
+	if err != nil {
+		return "", fmt.Errorf("error while geting subdomain for system URL: %s", err)
+	}
+	systemURL := strings.Replace(gatewayURL, subDomain, SystemSubDomain, -1)
+	return systemURL, nil
+}
+
+func getSubDomain(URL string) (string, error) {
```
alexellis: Is this in the Url struct already like `Host` or similar? Check the Go standard library for everything that is available. https://golang.org/pkg/net/url/

gitlab-status/handler.go
```diff
@@ -112,3 +135,88 @@ func validateRequest(req []byte) (err error) {
...
+	if targetURLerr != nil {
+		return aviableTargetURL, fmt.Errorf("error while formatting endpoint URL: %s", targetURLerr.Error())
+	}
+
+	aviableTargetURL[dashboardURL], targetURLerr = formatDashboardURL(gatewayURL, event)
+	if targetURLerr != nil {
+		return aviableTargetURL, fmt.Errorf("the dashboard url failed: %s", targetURLerr.Error())
+	}
+
+	return aviableTargetURL, nil
```
alexellis: aviable -> available (typo)

gitlab-status/handler.go
```diff
@@ -112,3 +135,88 @@ func validateRequest(req []byte) (err error) {
...
+		targetURL = aviableURL[logsURL]
+	}
+	if functionContext == sdk.StackContext &&
+		functionContext != eventService {
+		targetURL = aviableURL[dashboardURL]
+	}
+	return targetURL
+}
+
+func getTargetURLs(gatewayURL string, event *sdk.Event) (map[string]string, error) {
```
alexellis: Struct would be better for this for maintenance.

gitlab-status/handler.go
```diff
@@ -112,3 +135,88 @@ func validateRequest(req []byte) (err error) {
...
+	subdomain := strings.Split(parsedURL.Host, ".")
+
+	//Host is www.world.org and subdomain would be www aka. 0th element of the slice
+	return subdomain[0], nil
+}
+
+func setTargetURL(functionStatus, functionContext, eventService string, aviableURL map[string]string) string {
+	targetURL := aviableURL[endpointURL]
+	if functionStatus != sdk.StatusSuccess {
+		targetURL = aviableURL[logsURL]
```
alexellis: (typo on `aviable` again.)

alexellis:
Is there anything wrong with GitHub status? Can we check it too via the public instance of OpenFaaS Cloud? cc @ivanayov 

Do you need to edit stack.yml or gitlab.yml versions too?

gitlab-status/handler.go
martindekov: I actually have test for the case [here](https://github.com/openfaas/openfaas-cloud/blob/fb8e73e9d3a13493adbda3f224e35aed08debb9c/gitlab-status/handler_test.go#L324)

gitlab-status/handler.go
martindekov: I have checked that, and the `net/url` does not provide with something like subdomain, since it is subjective (`github.com`)

gitlab-status/handler.go
martindekov: fixed the typo

gitlab-status/handler.go
martindekov: Did the struct now

gitlab-status/handler.go
martindekov: fixed the `aviable` typos

gitlab-status/handler.go
ivanayov: What will happen if `gateway_public_url` is configured without subdomain and looks like `some.path.with-dots`?
Is this possible use-case?
If yes, we would like only to prefix the url, instead of replacing.

gitlab-status/handler.go
```diff
@@ -32,8 +40,22 @@ func Handle(req []byte) string {
 		log.Printf("error while building base URL to the API: %s", urlErr.Error())
 	}
 
+	gatewayURL := os.Getenv("gateway_public_url")
+	if gatewayURL == "" {
+		log.Printf("empty gateway_public_url unable unable to construct logs and endpoint URLs")
```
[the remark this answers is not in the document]
martindekov: Thanks for pointing that out 👍

gitlab-status/handler.go
```diff
@@ -112,3 +135,94 @@ func validateRequest(req []byte) (err error) {
...
+
+	availableTargetURL.DashboardURL, targetURLerr = formatDashboardURL(gatewayURL, event)
+	if targetURLerr != nil {
+		return availableTargetURL, fmt.Errorf("the dashboard url failed: %s", targetURLerr.Error())
+	}
+
+	return availableTargetURL, nil
+}
+
+type EndpointURL struct {
```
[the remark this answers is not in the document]
martindekov: Sounds good 👍

gitlab-status/handler.go
martindekov: I think and hope the structure of subdmain.o5s.io that way is a convention already, also we use the same convention when we have pretty_url for every user, I think people would like to have full cloud functionality when deploying it and setting 3 items separated with dots is the way, additional is that we name the functions from the cloud which are exposed in the repo has naming system-gitlab-event instead of gitlab-event so the router can make use of the subdomain system. And I think I saw somewhere auth.system as variable replacing system in the public URL.

martindekov: This ^ is just to show the comment as answer to the question above

gitlab-status/handler.go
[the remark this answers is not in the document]
martindekov: Done ✔️

gitlab-status/handler.go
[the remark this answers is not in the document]
martindekov: Amended that ✔️

martindekov:
Thanks for the suggestions @ivanayov I amended the changes that you suggested

martindekov:
This is rebased and the methods are extracted in the sdk in the newest commit https://github.com/openfaas/openfaas-cloud/pull/329/commits/96dbbaba53e9cc9a888148cac675d8a45ca6aea8

martindekov:
Once this merged/released if there are no further comments I will continue working on #356

gitlab-status/handler.go
alexellis: @ivanayov are you ok with this?

sdk/url_builders.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+const (
+	SystemSubDomain = "system"
+)
+
+func FormatEndpointURL(gatewayURL string, event *Event) (string, error) {
```
alexellis: Hi @martindekov I see duplication here.. can we do the SDK PR first then re-vendor rather than having a much bigger PR with duplication?

sdk/url_builders.go
```diff
@@ -0,0 +1,63 @@
...
+func FormatDashboardURL(gatewayURL string, event *Event) (string, error) {
+	systemURL, formatErr := FormatSystemURL(gatewayURL)
+	if formatErr != nil {
+		return "", fmt.Errorf("error while formatting dashboard URL: %s", formatErr.Error())
+	}
+
+	return fmt.Sprintf("%s/dashboard/%s", systemURL, event.Owner), nil
+}
+
+func GetSubDomain(URL string) (string, error) {
```
alexellis: `GetSubDomain` -> `GetSubdomain`

martindekov:
Ok @alexellis will try to do it by EOD along with cli change

martindekov:
Vendored in `0.8.8` I think this is ready for merge 👍

alexellis:
Please can you put it into the project view?

martindekov:
Yep 👍

alexellis:
I'm happy to merge this @martindekov. Are you confident there is no regression for GitHub or GitLab?

martindekov:
This is not ready, it redirects to `o6s.io` rather than my instance, the redirecting functionality works as expected, but does not point to the right place, I will need to fix this

martindekov:
Hey @alexellis I was wrong, was applying wrong `gateway_config.yml` file ... The change is working tested it e2e on my GitLab instance. If you want go ahead ad try it out I will re-base due to conflicts and it is ready to be merged

alexellis:
@martindekov do we still need this change? Does it need anything else other than the outstanding rebase?

Alex

martindekov:
Alistair did multiple changes on the dashboard so I will need to rebase and test this. As per my last comment this was ready to be merged when I left it, again it might not be that way now.

Waterdrips:
@martindekov the `log` endpoint from the dashboard has moved to `build-log`
this probably needs updating in this function too

alexellis:
Closing as inactive, if we still want to do this work, let's raise a new issue to discuss. Thanks for the contribution.

PR #331, opened 2018-11-14
Make validation turn on by default

### LINKED ISSUE #316: Make validation on by default for hmac/customers
## Expected Behaviour

If I mis-configure my system or leave a line out, I still want a secure experience that validates HMAC and the customers list. I.e. they should be on by default (even if no env/YAML setting is et)

So we can still turn these off, but by explicitly adding `validate_customers: false` etc into the YAML file. This means that you as a user/operator are deliberately choosing to turn it off.

## Current Behaviour

If I comment out the line for these values they get turned off.

## Possible Solution

* Update in SDK if necessary then re-release / re-vendor.
* Update consuming code and services
* Update docs on turning off

## Testing

Test that the validation is present for each function when the settings are commented out in the files.

The `faas-cli` has a new method to sign function bodies too (which we can use for the positive test)

martindekov:
Turn on validation on customers and hmac by default
now you must be explicit on turning it off by setting it
to `0` or `false` started by @ivanayov 

## Description

Before users had to set up explicitly the validation to `true` now the logic is turned around and you need to be explicit on turning it off by saying `false` to the variables `validate_hmac` and `validate_customers`

Closes #316 

## How Has This Been Tested?

Function added in sdk along with unit test

N.A. For users which have their validation removed by commenting out the variable they need to set the variable to `false` or `0` because commented out means true now

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/github-event/handler.go
+++ b/github-event/handler.go
@@ -39,7 +39,7 @@ func Handle(req []byte) string {
 		return fmt.Sprintf("%s cannot handle event: %s", Source, eventHeader)
 	}
 
-	if readBool("validate_customers") {
+	if ValidateCustomers() {
 		customersURL := os.Getenv("customers_url")
 
 		customers, getErr := getCustomers(customersURL)
@@ -91,8 +91,7 @@ func Handle(req []byte) string {
 		eventHeader == "installation_repositories" ||
 		eventHeader == "integration_installation" {
 
-		shouldValidate := os.Getenv("validate_hmac")
-		if len(shouldValidate) > 0 && (shouldValidate == "1" || shouldValidate == "true") {
+		if HmacEnabled() {
 			webhookSecretKey, secretErr := sdk.ReadSecret("github-webhook-secret")
 			if secretErr != nil {
 				return secretErr.Error()
@@ -323,4 +322,18 @@ func readBool(key string) bool {
 		return val == "true" || val == "1"
 	}
 	return false
+}
+
+func HmacEnabled() bool {
+	if val, exists := os.LookupEnv("validate_hmac"); exists {
+		return val != "false" && val != "0"
+	}
+	return true
+}
+
+func ValidateCustomers() bool {
+	if val, exists := os.LookupEnv("validate_customers"); exists {
+		return val != "false" && val != "0"
+	}
+	return true
 }
--- a/github-push/handler_test.go
+++ b/github-push/handler_test.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/openfaas/openfaas-cloud/sdk"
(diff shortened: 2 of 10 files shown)


github-push/handler_test.go
```diff
@@ -55,21 +55,3 @@ func Test_Handle_IssueComment(t *testing.T) {
 		t.Fail()
 	}
 }
-func Test_Handle_ValidateCustomers_Matched(t *testing.T) {
```
martindekov: Removed this test because It's intent is to test customers from URL but it actually gives error connected with secret, which I think should not be the case.

alexellis: Can you restore it please?

import-secrets/handler.go
```diff
@@ -157,3 +153,10 @@ type SealedSecret struct {
 	Metadata   *metav1.ObjectMeta `yaml:"metadata"`
 	Spec       SealedSecretSpec   `yaml:"spec"`
 }
+
+func readBool(key string) bool {
```
alexellis: Can you reference this from the sdk package instead?

github-status/handler.go
```diff
@@ -33,8 +33,7 @@ var (
 
 // Handle a serverless request
 func Handle(req []byte) string {
-
-	if hmacEnabled() {
```
alexellis: I preferred the old name because it's more descriptive and abstracts away the env-var. Can you revert please?

github-push/handler_test.go
martindekov: Ok no problems

github-status/handler.go
martindekov: Yep no problems

import-secrets/handler.go
martindekov: I thought about that but the function is not in the sdk package so it will fail, maybe when this is merged and package bumped up we can make small pr referencing it from the SDK

github-event/handler.go
```diff
@@ -320,7 +319,7 @@ func getCustomers(customerURL string) ([]string, error) {
 
 func readBool(key string) bool {
```
alexellis: This should take an override like `readBool(key string, defaultVal bool)` otherwise it's not the equivalent of `readBool` elsewhere in OpenFaaS where null/empty meant false. Alternatively change `readBool` to something else like a more specific validateHmac() or hmacEnabled().

martindekov: There is `HmacEnabled()` in SDK, should I create `HmacDisabled()` or change the logic of `HmacEnabled()`? Which way should I go with ?

martindekov:
@alexellis need input on this also left a comment

martindekov:
It is done ✔️

github-event/handler.go
martindekov: Done ✔️

alexellis:
Please remove the HmacEnabled methods from each handler and point at the new SDK with vendoring - I'll do a release now for you.

martindekov:
Yeah will open PR with the change

PR #333, opened 2018-11-15
Add annotations to list-functions

martindekov:
Annotation support for functions was added in previous
commits to the dashboard and buildshiprun but was forgotten
for list-functions and this change now enables it

## Description

Small fix so list-functions now appends annotations so the actual annotation returned from the function

## How Has This Been Tested?

In end to end test on kubernetes instance

They will be able to see their URLs depending on whether or not they have GitLab or GitHub Cloud instances

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/list-functions/handler.go
+++ b/list-functions/handler.go
@@ -86,4 +86,5 @@ type function struct {
 	InvocationCount float64           `json:"invocationCount"`
 	Replicas        uint64            `json:"replicas"`
 	Labels          map[string]string `json:"labels"`
+	Annotations     map[string]string `json:"annotations"`
 }


alexellis:
OK, so now that we've added them to the serialisation.. do we need to check them for any reason or just return them to the client? 👍

alexellis:
Thanks for this. We now need a PR to bump the version in the two .YAML files where list-functions is mentioned before I can rebuild/republish. @martindekov

martindekov:
> OK, so now that we've added them to the serialisation.. do we need to check them for any reason or just return them to the client? 👍

I guess we can check the annotations for safety, but haven't thought on that

PR #335, opened 2018-11-15
Add documentation fix on minio

martindekov:
Fixing details around minio set up because the documentation
was little big confusing and incorrect.

## Description

The part of the documentation which describes how to set up minio had some misleading information for example `name` attribute was saying that it should be set to the DNS where minio could be accessed from inside the cluster

## How Has This Been Tested?

N.A.

Easier set up for minio in case you have no idea how services, namespaces etc. work on creating DNS to access service

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -355,14 +355,33 @@ kubectl create secret generic -n openfaas-fn \
 Install Minio with helm
 
 ```
-helm install --name cloud --namespace openfaas \
+helm install --name cloud-minio --namespace openfaas \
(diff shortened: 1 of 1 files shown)


alexellis:
Looks like the name they picked in the helm chart was changed since the docs were published.

Do you need to update any .yml files with this too?

martindekov:
I don't think it is necessary since the only problem I had was with naming ^

Issue #325, opened 2018-10-31
Dashboard: ship dist folder in final stage of Docker build


alexellis:
## Expected Behaviour

Dashboard: ship dist folder in final stage of Docker build

## Current Behaviour

We do `make` which takes `src` into a Node container, runs web pack and emits `dist` folder which is then built into a new container. We can streamline process and avoid any permissions issues.

alexellis:
I'm pretty sure we'll have to stop using a function deployment type and start using a custom Dockerfile.

@bartsmykla please use `faas-cli build --shrinkwrap` then base the code on what ends up in the "build/function-name" folder.

bartsmykla:
@burtonr sorry for not responding to you, but I missed the message, before I started working on it. I created PR for that change: https://github.com/openfaas/openfaas-cloud/pull/338 I would be very grateful if you could look on it with @alexellis and give your opinions

PR #338, opened 2018-11-21
Moved step of building static assets to dockerfile

bartsmykla:
Closes: https://github.com/openfaas/openfaas-cloud/issues/325

## Description

- Changed dashboard-function type to dockerfile and moved building
  static files step into it. There is no need to have dist directory
  inside repository now

## How Has This Been Tested?

- Deployed dashboard to my cluster and tested if all features works
- Started development version of dashboard `yarn start` inside `dashboard/client` directory checking if everything works for development purposes also

Users should not be affected because only building process of dashboard changed. The results should be the same.

## Checklist:

I have:

- [ ] added unit tests


alexellis:
Thanks for getting to this.

What's the new instructions for local development vs publishing a container? Should we give some help to new contributors?

bartsmykla:
Actually I removed mention of manually step of building it using `make build` from documentation, and nothing else changed. Now for deploying a container you can use just `faas up` and it will build everything and deploy automatically.

Local development didn't change at all. I just tested if everything about it works as it was before.

dashboard/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+
+### Final stage
+
+FROM node:10.12.0-alpine
+
+WORKDIR /root/
+
+RUN addgroup -S app && adduser -S -g app app
+
+RUN apk --no-cache add curl vim \
```
ewilde: is there a reason this `apk` install / `curl` isn't done in the previous build stage and then copied into this final build stage

alexellis:
Please can you review the instructions for developers to make sure there is adequate help for getting started and running locally / in a Docker image.

alexellis:
Needs a rebase please..

dashboard/Dockerfile
alexellis: @bartsmykla WDYT?

bartsmykla: Hi @ewilde huge sorry, for not getting back to you. I'll respond to you today!

bartsmykla:
It's rebased now

bartsmykla:
But I need to test it again, after rebasing with master

alexellis:
(OK so this is with you for now.)

bartsmykla:
@alexellis rebased again, and tested. I'm confident that it's working as expected

dashboard/Dockerfile
bartsmykla: @ewilde I just understood your point, and I fixed that. Thank you, and sorry for getting back to this so late

bartsmykla:
Derek remove label: blocked

alexellis:
@Waterdrips do you feel we could we benefit from continuing to implement this change? The primary difference is that the assets are not generated on disk and built into a container, they're built directly within the container we ship.

Waterdrips:
So I personally feel that adding over 150 lines just to remove the need to run `make && faas-cli up` instead of `faas-cli up` isnt worth it.

Plus we then have to maintain these templates rather than just change/upgrade using one of the other openfaas templates

Issue #340, opened 2018-11-22
UI Dashboard: Change reading index.html to async


bartsmykla:
Referring to this comment: https://github.com/openfaas/openfaas-cloud/pull/295#pullrequestreview-175725899 it would be good to change `readFileSync` to `readFile`
## Expected Behaviour
file `index.html` should being read asynchronously

## Current Behaviour
file `index.html` is being read synchronously

alexellis:
Derek add label: design/approved

bartsmykla:
Derek assign: me

PR #348, opened 2018-11-27
Changed readFileSync to readFile inside dashboard handler

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/issues/340

## Description
Files before serve are read asynchronously now.

## How Has This Been Tested?
I have deployed dashboard to cluster on Digital Ocean and test if every endpoint works.

Users should not be impacted, because except simple change from sync to async version of reading files before serve results need to be the same.

## Checklist:

I have:

- [ ] added unit tests


alexellis:
@bartsmykla is the build stuck?

Issue #236, opened 2018-09-26
UI: Dashboard - introduce blank slate feature


alexellis:
## Expected Behaviour

When you land on the page there should be a link or action button clearly displayed when you have no functions. This is the blank slate / state for the dashboard. See the regular OpenFaaS UI portal for an example.

## Current Behaviour

We see a vague error message like "No functions", but probably don't need the grid rendered at all. There is no clear next-action for the user to take and we can improve that.

![blank-slate](https://user-images.githubusercontent.com/6358735/46096974-8687f380-c1b8-11e8-9a9d-d5f05ef55f5e.png)

## Possible Solution

We should make some suggestions / designs of how we can improve this. Allen cc @allendjal could probably help here.

allendjal:
Proposed mockup attached here. Includes CTAs: create a new function + learn about openfaas functions
![functions-blank-slate](https://user-images.githubusercontent.com/38286758/46130827-16e12b00-c1ef-11e8-8d86-a2e8f57620b0.jpg)

alexellis:
@bartsmykla how many hours do you think this task would take you to complete?

Thanks,

Alex

bartsmykla:
4h

Waterdrips:
Derek assign: me

PR #349, opened 2018-11-29
Improved empty state when user has no functions

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/issues/236

## Description
Full version:
image

Mobile version:
image

## How Has This Been Tested?

I have tested it by deploying it to my cluster and checked
if every state is working (when there are functions,
when there are no functions, when functions are loading)

Current users if want to see the new empty state, they need to update dashboard and redeploy it to the newest version.

## Checklist:

I have:

- [ ] added unit tests


bartsmykla:
@allendjal @jaigouk what do you think?

alexellis:
Where will the two links land?

alexellis:
Tutorials -> workshop ?

alexellis:
Create a function probably needs to go to the OpenFaaS Gist for the community cluster.

alexellis:
Needs a rebase now and the link suggestions so that the first one is specific to OpenFaaS Cloud to prevent confusion.

bartsmykla:
@alexellis do we have that gist ready, or it needs to be prepared first?

alexellis:
You know the gist for the community cluster right? It's in #openfaas-cloud pinned as the title of the channel - it's the doc everyone is using to get started in the shared environment.

alexellis:
@bartsmykla rebase required now.

bartsmykla:
@alexellis link changed and branch rebased

alexellis:
Thanks for the rebase. Let's check the links over before merging. What are they pointing at now?

alexellis:
I think we need to work on where the links are pointing to - to be specifically about OpenFaaS Cloud. The second logo also needs to change to not be the same thing because this confusing. (advice from Allen)

bartsmykla:
@alexellis so as we discussed I can change the second image to be the same as the first one, and do we have any suggestions which links do we want to to?

alexellis:
Released via linked PR

Issue #347, opened 2018-11-27
[Feature] Show and filter functions from my orgs in my dashboard


alexellis:
## Story

As an OpenFaaS Cloud user visiting my dashboard I want to be able to see my functions plus all the functions belonging to my organisations so that I can collaborate with my team.

This related issue #234 shows where we started to store the claims in the user's JWT token. This can be read in the back-end function of the dashboard and be used to call the `list-functions` function with all the `usernames` vs only the logged in user.

I.e. Alex Ellis (alexellis) is a member of vmware and openfaas, therefore call list-functions for: `[alexellis, openfaas, vmware]`.

## Acceptance criteria (priority order)

1. Given I am a member of organisation X, I can see functions belonging to organisation X in my dashboard Addressed by: #289
2. The sort-order shows my functions first, followed by each org ordered by name using string comparisons
3. I can filter based upon `owner` / organisation using a drop-down in the UI
4. My filter selection is "remembered" / stored in a local cookie.

ivanayov:
Derek assign: me

Waterdrips:
Derek assign: me

alexellis:
Requirements for 1) closes: #289

alexellis:
Requirements for 2) 3) and 4) are closed by: #361

PR #353, opened 2018-12-05
Show functions from organizations

ivanayov:
## Description

With this change users will be able not only to see functions they
own, but also the ones that belong to the organizations they are
members of

Closes 1. and 2. from #347 

## How Has This Been Tested?

Deployed the dashboard + list-functions on a cluster.
Tested with `docwareiy/of-cloud-dashboard:0.3.1-list-orgs-fn`
After setting a cookie of type `openfaas_cloud_token=<jwt with organizations>` the listed functions were functions from the user + functions from orgs.

After deleting the cookie, listed functions were only functions from user.

Tested with OAuth it's listing functions from both user and orgs, sorting the user first.

N/A

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required) N/A
- [ ] added unit tests

### DIFF
--- a/dashboard/of-cloud-dashboard/handler.js
+++ b/dashboard/of-cloud-dashboard/handler.js
@@ -6,6 +6,10 @@ const request = require('request');
 module.exports = (event, context) => {
   const { method, path } = event;
 
+  parseOrganizations: parseOrganizations;
+  decodeCookie: decodeCookie;
+  getCookie: getCookie;
+
   if (method !== 'GET') {
     context.status(400).fail('Bad Request');
     return;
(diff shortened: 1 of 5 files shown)


alexellis:
Unfortunately we can't read the cookie on the client side, as it's set to HTTP only. We need to move this JavaScript code into the dashboard backend which is a node function.

Here's an example of how we've done this with the logout functionality with @bartsmykla 

of-dashboard-cookie-claims

ivanayov:
@alexellis the code is now moved to the backend and tested with OAuth

alexellis:
That's awesome!

Do we have an issue to show the username as a box with the initials i.e. [OF] / [AE] for when there is more than one org+user?

ivanayov:
Yes, it's #359

dashboard/of-cloud-dashboard/handler.js
```diff
@@ -62,15 +66,49 @@ module.exports = (event, context) => {
...
+var parseOrganizations = function (cookie) {
+  var decodedCookie = decodeCookie(cookie);
+  console.log("decodedCookie -> ", decodedCookie);
+  if (decodedCookie && 'organizations' in decodedCookie) {
+    return decodedCookie.organizations;
+  }
+  return '';
+}
+
+var atob = function atob(str) {
```
alexellis: What else could we call `atob`? `decodeBase64`?

Issue #354, opened 2018-12-06
gitlab-status refactoring on http.Client


alexellis:
## Expected Behaviour

Use default client or create client once and re-use

## Current Behaviour

Creates multiple HTTP clients one for each function in stack.

```
client := &http.Client{}
```

## Possible Solution

Refactor/reduce.

PR #364, opened 2018-12-17
Moved creation of http client outside of sendReport

bartsmykla:
Closes https://github.com/openfaas/openfaas-cloud/issues/354

## Description
http.Client is created in root scope, so it's not recreated
every time, when sendReport is being invoked.

## How Has This Been Tested?
Deployed to my cluster on AWS.

They shouldn't be because the result should be the same, just performance should be improved.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/gitlab-status/handler.go
+++ b/gitlab-status/handler.go
@@ -88,8 +88,8 @@ func sendReport(URL string, token string, state string, desc string, context str
 		return fmt.Errorf("error while creating request to GitLab API: %s", reqErr.Error())
 	}
 	req.Header.Set("PRIVATE-TOKEN", token)
-	client := &http.Client{}
-	resp, clientErr := client.Do(req)
+
+	resp, clientErr := http.DefaultClient.Do(req)
 	if clientErr != nil {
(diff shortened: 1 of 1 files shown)


gitlab-status/handler.go
```diff
@@ -88,8 +90,8 @@ func sendReport(URL string, token string, state string, desc string, context str
 		return fmt.Errorf("error while creating request to GitLab API: %s", reqErr.Error())
 	}
 	req.Header.Set("PRIVATE-TOKEN", token)
-	client := &http.Client{}
```
alexellis: Can you use `http.DefaultClient` instead?

bartsmykla: Done

bartsmykla:
Done

gitlab-status/handler.go
```diff
@@ -77,6 +77,8 @@ func appendParameters(URL string, state string, desc string, context string) (st
 
 }
 
+var httpClient = http.DefaultClient
```
alexellis: No need for this pointer. Please just inline it.

bartsmykla: Done

PR #377, opened 2019-01-16
mario-mazo

mario-mazo:
mario-mazo
Signed-off-by: Mario Mazo <buster.mazo@gmail.com>


mario-mazo:
I would like to try it to see if it makes sense to deploy it at work later on.

initially I would like to try it with  grant/deny access to my private aws env. Like a super cheap vpn

PR #385, opened 2019-01-29
Update NetworkPolicy

bartsmykla:
This issue need to be resolved before: https://github.com/openfaas/faas-netes/issues/367

## How Has This Been Tested?

I deployed OpenFaaS Cloud on new cluster created on 1 droplet (8GB RAM, 8CPU cores) with `calico` network plugin, then I applied network policies and tried to connect from:
- `openfaas-fn` namespace from functions with label `role=openfaas-system` to functions inside `openfaas` namespace
- from functions in `openfaas` namespace to functions in `openfaas` namespace
- from functions in `openfaas` namespace to functions in `openfaas-fn` namespace
- checked if all deployed functions work from the internet
- checked if dashboard is working

They shouldn't be if they are not using NetworkPolicies (they were not working correctly before, and nobody from the community reported any problems, so I assume that nobody was using it).

If somebody want to use NetworkPolicies now, there is only need to deploy it do the cluster using command:
```
kubectl apply -f ./yaml/network-policy
```

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -231,13 +231,6 @@ kubectl apply -f ./yaml/core
 kubectl apply -f ./yaml/network-policy
 ```
 
-(Optional) Add a role of "openfaas-system" using a label to the namespace where you deployed Ingress Controller. For example if Ingress Controller is deployed in the namespace `ingress-nginx`:
-```
-kubectl label namespace ingress-nginx role=openfaas-system
-```
-
-If you don't have Ingress Controller installed in cluster. [Read this](#troubleshoot-network-policies)
-
 #### For Swarm
 
 Create the secret for your registry
@@ -469,23 +462,12 @@ kubectl get events --sort-by=.metadata.creationTimestamp -n openfaas-fn
 ```
 
 ##### Troubleshoot Network Policies
-The NetworkPolicy configuration is designed to work with a Kubernetes IngressController. If you are using a NodePort or LoadBalancer you have to deploy NetworkPolicy below.
-```yaml
-apiVersion: networking.k8s.io/v1
-kind: NetworkPolicy
-metadata:
-  name: gateway
-  namespace: openfaas
-spec:
-  policyTypes:
-    - Ingress
-  podSelector:
-    matchLabels:
-      app: gateway
-  ingress:
-    - from: []
-      ports:
-        - port: 8080
+
+The NetworkPolicy configuration is designed to work with a namespaces `openfaas` and `openfaas-fn`.
+If your are using different namespaces you need to make sure that they're labeled correctly:
+```
+kubectl label ns openfaas role=openfaas-system
+kubectl label ns openfaas-fn role=openfaas-fn
 ```
 
 #### Troubleshoot Swarm
--- a/yaml/network-policy/ns-openfaas-net-policy.yml
+++ b/yaml/network-policy/ns-openfaas-net-policy.yml
@@ -8,7 +8,14 @@ spec:
     - Ingress
   podSelector: {}
   ingress:
-    - from:
-        - namespaceSelector:
-            matchLabels:
-              role: openfaas-system
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          role: openfaas-system
+
+    - namespaceSelector:
+        matchLabels:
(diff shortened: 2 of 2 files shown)


docs/README.md
```diff
@@ -469,23 +462,12 @@ kubectl get events --sort-by=.metadata.creationTimestamp -n openfaas-fn
...
-      app: gateway
-  ingress:
-    - from: []
-      ports:
-        - port: 8080
+
+The NetworkPolicy configuration is designed to work with a namespaces `openfaas` and `openfaas-fn`.
+If your are using different namespaces you need to make sure that they're labeled correctly:
+```
+kubectl label ns openfaas role=openfaas-system
```
alexellis: Are these labels in the namespaces.yaml file in the faas-netes repo? If not, could we move them there instead of applying the labels manually?

alexellis: https://github.com/openfaas/faas-netes/blob/master/namespaces.yml

stefanprodan: I think it's an example for custom namespaces, we could use different names

bartsmykla: It's like @stefanprodan said

stefanprodan:
I've tested the network policies on GKE by creating the two namespaces and running curl pods in both of them. I can confirm that pods without the `openfaas-system` label can't access the openfaas namespace. My concern is that if you use an ingress controller (that sits in a different namespace), it will not be able to reach the gateway.  @bartsmykla have you tested this with nginx or any other ingress controller?

bartsmykla:
I have tested it with `nginx-ingress`, it was working as expected.

stefanprodan:
The nginx ingress was deployed in the kube-system namespace or in the openfaas namespace?

bartsmykla:
@stefanprodan `kube-system`

stefanprodan:
Something must be wrong then, the netowrk policy should not allow traffic to the `openfaas `namespace unless the caller is in the `openfaas-fn` namespace and has the `openfaas-system` label.

docs/README.md
```diff
@@ -469,23 +462,12 @@ kubectl get events --sort-by=.metadata.creationTimestamp -n openfaas-fn
...
-spec:
-  policyTypes:
-    - Ingress
-  podSelector:
-    matchLabels:
-      app: gateway
-  ingress:
-    - from: []
-      ports:
-        - port: 8080
```
stefanprodan: @bartsmykla this policy would allow an ingress controller to reach the gateway on 8080, without it I don't see how the nginx can forward traffic to the gateway.

bartsmykla: I think I just annotated `kube-system` with `openfaas-system` label

bartsmykla: But it was a month ago, so I'm sure in 80% only.

stefanprodan: @alexellis do you see any issues if we allow access to the 8080 port from any namespace? I'm guessing the basic auth is turned on so it shouldn't be a problem. I don't like the idea of labeling `kube-system`.

bartsmykla: I think we should ask @alexellis - I don't have any problem with it.

alexellis: You could DoS the core services from a function if we allowed access.

If we didn't apply the label, what would this cover?

stefanprodan: We could allow traffic from nginx using a pod selector:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: openfaas
  namespace: openfaas
spec:
  policyTypes:
    - Ingress
  podSelector: {}
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          role: openfaas-system
    - namespaceSelector:
        matchLabels:
          role: openfaas-fn
      podSelector:
        matchLabels:
          role: openfaas-system
    - podSelector:
        matchLabels:
          app.kubernetes.io/name: ingress-nginx
```

alexellis:
Derek set title: Update NetworkPolicy

bartsmykla:
Guys. I'm closing PR, because of it's age. Feel free to recreate it.

matipan:
Here is a new PR for the network policies: https://github.com/openfaas/openfaas-cloud/pull/455

PR #387, opened 2019-01-31
Update auth README

ivanayov:
The external redirect domain was previoulsy pointing on port 8081, which is not applicable when a domain name is used

Raised on slack after testing OAuth with orc-bootstrap


alexellis:
Thanks

PR #390, opened 2019-02-06
add user rajatjin to customers list

rajatjin:
Hi Alex,

can you please approve 'rajatjin' account to be allowed in the customers list.

Thanks
Rajat Jindal


alexellis:
Sure. Why do you have two accounts?

Issue #391, opened 2019-02-06
Task: validate CUSTOMERS file before use


alexellis:
## Expected Behaviour

We should validate all entries in the CUSTOMERS file before use - to prevent anyone getting their functions overriden by accident. See issue #314 

## Possible Solution

Since this is relying on a human element of not merging a PR to the customers file with two potentially clashing usernames, we should implement a unit test and sdk method in the sdk of OpenFaaS Cloud that can validate a list of usernames.

I.e.

OK 

```
alexellis
stefan
ivana
```

Not OK 

```
alexellis
alexellis-suffix
stefan
ivana
```

Not OK 

```
alexellis
stefan
stefan-prodan
ivana
```

OK 

```
alexellis
stefan-prodan
ivana
```

alexellis:
The PR should be raised here: https://github.com/openfaas/openfaas-cloud/tree/master/sdk

alexellis:
No need to test end-to-end, just via unit tests.

https://blog.alexellis.io/golang-writing-unit-tests/

rajatjindal:
how about

```
alexellis   
stefan-prodan  
stefan  
ivana
```

and

```
alexellis
alexe
stefan
ivana
```

alexellis:
What about it?

alexellis:
alexe won't clash because of the separator being `-`

alexellis:
`alexe` => `alexe-my-function`
`alexellis` => `alexellis-my-function`

martindekov:
Derek close: fixed in #393

PR #393, opened 2019-02-06
Add validation customer names list

viveksyngh:

## Description

 #391

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/sdk/customers.go
+++ b/sdk/customers.go
@@ -1,6 +1,9 @@
 package sdk
 
-import "os"
+import (
+	"os"
+	"strings"
+)
 
 // ValidateCustomers checks environmental
 // variable validate_customers if customer
@@ -9,5 +12,21 @@ func ValidateCustomers() bool {
 	if val, exists := os.LookupEnv("validate_customers"); exists {
 		return val != "false" && val != "0"
 	}
+	return true
+}
+
+//ValidateCustomerList validate customer names list
+func ValidateCustomerList(customers []string) bool {
+	for i, customerName := range customers {
+		for j, cn := range customers {
+
+			if i != j {
+				if strings.HasPrefix(cn, customerName+"-") {
+					return false
+				}
+			}
+		}
+	}
+
 	return true
 }
--- a/sdk/customers_test.go
+++ b/sdk/customers_test.go
@@ -53,3 +53,44 @@ func Test_ValidateCustomers(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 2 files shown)


sdk/customers.go
```diff
@@ -11,3 +14,19 @@ func ValidateCustomers() bool {
...
 }
+
+//ValidateCustomerList validate customer names list
+func ValidateCustomerList(customers []string) bool {
+	for i, customerName := range customers {
+		for j, cn := range customers {
+
+			if i != j {
+				if strings.HasPrefix(cn, customerName+"-") {
+					return false
```
rajatjindal: i think it should also return a error msg specifying which username(s) caused this validation to fail and if possible, we should expose that in the PR automatically.

sdk/customers.go
```diff
@@ -11,3 +14,19 @@ func ValidateCustomers() bool {
 	}
 	return true
 }
+
+//ValidateCustomerList validate customer names list
```
rajatjindal: where are we calling this function?

sdk/customers.go
```diff
@@ -11,3 +14,19 @@ func ValidateCustomers() bool {
...
+func ValidateCustomerList(customers []string) bool {
+	for i, customerName := range customers {
+		for j, cn := range customers {
+
+			if i != j {
+				if strings.HasPrefix(cn, customerName+"-") {
```
rajatjindal: while we are at it, do we want to add validation for duplicate check?

sdk/customers.go
viveksyngh: This will be called inside function once we release new version of sdk and the vendor it inside the function.

sdk/customers.go
alexellis: That's a good point. Perhaps we can change this to return the names which conflict @viveksyngh for debugging and better diagnostics?

alexellis:
I'll merge this as the initial implementation.

A list of conflicting users would be ideal for a follow-up PR.

alexellis:
@martindekov @viveksyngh @s8sg please advise on which functions you believe we need to add this to?

I'm guessing system-github-event and system-gitlab-event?

martindekov:
Yes, for GitLab the `system-gitlab-push` has no functionality for validating customers(it was intended). I had issue somewhere to leave only `system-*-event` functions to have those(hmac validation,customers ...) and depricate those from the `system-*-push` functions.

martindekov:
But isn't the `CUSTOMER` validation already added ?

Issue #392, opened 2019-02-06
Add Makefile for ci-armhf-build / ci-armhf-push


alexellis:
## Expected Behaviour

So that we can build and push images officially for armhf and in development we should implement a Makefile like the ones we did for other OF projects.

## Possible Solution

Cover:

* ./auth
* ./router
* ./of-builder

I don't believe that we need disparate Dockerfiles, multi-arch should work.

PR #394, opened 2019-02-06
Add ARMHF/ARM64 build image support

rdimitrov:
## Description
* All architectures can use the same Dockerfiles
* Added a script similar to `publish-arm.sh` that is used for building each component. This way it's easier to extend in future and it also takes into account the host architecture it's being run from.
* There are single `make ci-arm**-*` targets in the main Makefile that trigger a build for all configured components. Or you can hop into each folder and build it separately. 
* A simple `ci-arm-build` was going to be enough as well since the script is aware of the host arch, but decided to keep the same make style as we have so far. This will be helpful if at some point we decide to add this repo to the more general `publish-arm.sh` script and it will work out of the box.

Closes: https://github.com/openfaas/openfaas-cloud/issues/392
## How Has This Been Tested?

Tested if the images are building successfully.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/auth/Dockerfile
+++ b/auth/Dockerfile
@@ -1,5 +1,5 @@
 FROM golang:1.10-alpine AS builder
-
+RUN apk add --no-cache g++
 WORKDIR /go/src/github.com/openfaas/openfaas-cloud/auth
 
 COPY vendor     vendor
--- a/router/Dockerfile
+++ b/router/Dockerfile
@@ -1,5 +1,5 @@
 FROM golang:1.10-alpine AS builder
-
+RUN apk add --no-cache g++
 WORKDIR /go/src/github.com/openfaas/openfaas-cloud/router
 
 COPY main.go            .
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,20 @@
 TAG?=latest
-
+.PHONY: all
 all:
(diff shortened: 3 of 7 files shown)


auth/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM golang:1.10-alpine AS builder
-
+RUN apk add --no-cache g++
```
alexellis: Why do we need this?

router/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM golang:1.10-alpine AS builder
-
+RUN apk add --no-cache g++
```
alexellis: I can't think why we would need this?

rdimitrov: It fails while building the docker image otherwise

rdimitrov: That was happening only on ARM64 though.

zeerorg:
Should I add a `stack.armhf.yml` file so that the function image can be created and uploaded, along with this PR ?

alexellis:
Thank you for this PR, the work is a little out of sync now, and we have a newer mechanism for multi-arch, so I'm going to close in favour of a new Issue.

PR #399, opened 2019-02-09
Dashboard trailing slash url bug

zeerorg:
Signed-off-by: Rishabh Gupta<r.g.gupta@outlook.com>

## Description
when you open user dashboard with a trailing slash and go to a specific function deployed by the user the url is wrong and hence the function is not loaded correctly.

Duplicating the bug:

1. Open : https://system.o6s.io/dashboard/alexellis/
2. Click on any function and there will be a "Page not found"

## How Has This Been Tested?
Tested this on a arm deployment. http://system.o6s.zeerorg.site/dashboard/alexellis/ works fine since the patch has been applied.

The existing users are not impacted.

## Checklist:

I have:

- [ ] added unit tests


alexellis:
LGTM

PR #413, opened 2019-03-06
Add scottg489 to CUSTOMERS file

ScottG489:
## How did you find out about OpenFaaS Cloud?
I found out about it while google around for serverless solutions which use docker containers to represent functions. I've worked with docker extensively in the past and have also been interested in serverless for a while. Docker seemed like a perfect fit so I was surprised to find no on was using it this way. That is, until I came across this project :)

## What would you like to run there?
I would like experiment with it for small personal projects. I'm particularly interested in using it for CI as serverless seems like a perfect use case. I also have a few small personal projects I'd like to port to a serverless architecture.

### DIFF
--- a/CUSTOMERS
+++ b/CUSTOMERS
@@ -57,6 +57,7 @@ rimiti
 roncrivera
 rorpage
 s8sg
+scottg489
 stefanprodan
 tarunmangukiya
 telackey


alexellis:
Merged

PR #418, opened 2019-03-23
Add a new e2e test for openfaas cloud

matipan:
Signed-off-by: Matias Pan <matias.pan26@gmail.com>

## Description
This PR adds a new function that performs an end-to-end test of openfaas cloud. Fixes #403.

#### How does it work?
This function tests if OpenFaaS Cloud is correctly listening to github push events, building and re-deploying a function. To validate this the function does the following:
* Modify the source file of a Go function to respond with the current timestamp
* Call that function in a loop every 15 second until the function responds with the timestamp we expected

If it responded ok we return `true`. `false` is returned whenever the test failed or something bad happened.

#### Configure the function
@alexellis In order to put this function on the OpenFaas Cloud Community Cluster we need to configure the following parameters:
* **gateway**: gateway to the cluster
* **image**: name of the image of this function
* **functionOwner** and **functionRepo**: owner and repo of the function that it's being used to do the test
* **goFilePath**: the path to the Go file of the function,
* **functionURL**: URL where the function will be deployed,
* **applicationID**: This function authenticates with github as an application, once the application was installed(it needs permissions to write the contents of the repository) use the ID here,
* **secrets**: create two secrets on your OpenFaaS Cloud installation, one with the secret key of the github app and another one with the private key. Basically we need to do the same thing that was done on derek.

We also need to create a repository with a plain Go function and make sure that OpenFaaS Cloud is installed on that repository.

## How Has This Been Tested?

I tested this using my own installation of OpenFaaS Cloud on a GKE cluster. Whenever the result of the test was ok it responds with true:
ofc-test

## How are existing users impacted? What migration steps/scripts do we need?
None

## Checklist:

I have:

- [x] updated the documentation and/or roadmap (if required)
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [ ] added unit tests

PR #422, opened 2019-03-26
Add support for dockerfile template

### LINKED ISSUE #421: Support for Dockerfiles template in self-hosted clusters
## Expected Behaviour

We should make support available for the "dockerfile" template on self-hosted clusters.

## Current Behaviour

Unavailable see: #417 

## Possible Solution

* A flag is needed for OpenFaaS Cloud to enable use of the dockerfile template - validate this in the git-tar function
* This flag should be off by default
* Update git-tar's Dockerfile to use the latest faas-cli
* Test e2e

~~Explore what is missing in the code (probably the `faas-cli build --shrinkwrap` code is unavailable.~~

~~The `--shrinkwrap` option produces a directory which can be made into a tar for usage with an external Docker builder. On OFC this is Buildkit.~~

1. Deploy OFC
2. `faas-cli new --lang dockerfile fn`
3. `mv fn.yml stack.yml`
4. Commit to Git repo

## Context

This must be turned off by default so that it's not deployed into the Community Cluster.

matipan:

## Description
This PR adds support for the dockerfile template. In order to enable support users have to set `enable_dockerfile_lang` to true, it is false by default.
In order for this to be implemented the following changes were introduced:
* Update git-tar's faas-cli version to 0.8.6
* Modify git-tar to check if there are dockerfile functions and if the feature is enabled or not

Fixes #421 

## How Has This Been Tested?

I tested this using my own ofc installation and deploying [matipan/of-git-tar:dockerfileSupport](https://hub.docker.com/r/matipan/of-git-tar). Test cases:
* **If the feature is enabled and there is a dockerfile function**: [here](https://github.com/matipan/dtest/runs/84836104) you can see that the build and deploy was successful
* **If the feature is disabled and there is a dockerfile function**: [here](https://github.com/matipan/dtest/runs/84843643) you can see that the build failed with the error message: _detected a dockerfile function but feature is not enabled_
* **If the feature is enabled/disabled a function that it's not dockerfile can be successfully deployed**: I validated this using the e2e test from [this PR](https://github.com/openfaas/openfaas-cloud/pull/418)

Users will not be impacted since the feature is disabled by default.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/gateway_config.yml
+++ b/gateway_config.yml
@@ -34,6 +34,9 @@ environment:
   prometheus_port: 9090
   metrics_window: 60m
 
+  # Enable support for the dockerfile template
+  enable_dockerfile_lang: false
+
 # To use a shared Docker Hub account.
 #  repository_url: docker.io/ofcommunity/
 #  push_repository_url: docker.io/ofcommunity/
--- a/git-tar/Dockerfile
+++ b/git-tar/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.10.4-alpine3.7 as build
 RUN apk --no-cache add curl \ 
     && echo "Pulling watchdog binary from GitHub." \
     && curl -sSL https://github.com/openfaas/faas/releases/download/0.9.5/fwatchdog > /usr/bin/fwatchdog \
-    && curl -sSL https://github.com/openfaas/faas-cli/releases/download/0.7.3/faas-cli > /usr/local/bin/faas-cli \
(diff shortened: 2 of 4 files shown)


alexellis:
LGTM

alexellis:
LGTM

alexellis:
@matipan please can you pull in / re-sync any files from OFC into orc-bootstrap which are required for these changes? Thank you 👍

PR #430, opened 2019-04-07
Add user labels feature

alexellis:
Signed-off-by: Alex Ellis <alexellis2@gmail.com>

## Description

Allows users to pass their own values through stack.yml. These
still need to be consumed within buildshiprun and then forwarded
to the deployment request.

## How Has This Been Tested?

Tested with unit tests.

Tested end-to-end on live environment with:

* 1x function with a label override to prevent scaling to zero (got com.openfaas.scale.zero=false)
* 1x function without a label to prevent scaling to zero (got com.openfaas.scale.zero=true)

Sync the versions in ofc-bootstrap's stack.yml template so that users can benefit when they upgrade.

No immediate impact since faasIdler is off by default.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -25,6 +25,9 @@ const (
 	GitLab = "gitlab"
 )
 
+const scaleToZeroDefault = true
+const zeroScaleLabel = "com.openfaas.scale.zero"
+
 var (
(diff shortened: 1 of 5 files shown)


alexellis:
Please propose the HTTP client idea. I think `http.DefaultClient` is better than `http.Client{}`

alexellis:
Thank you for the review.

Issue #425, opened 2019-03-28
Feature: Support CI/CD from multiple branches to corresponding environments


sachin-dhamane:
We keep long term projects that have longer testing cycle on a different branch. We need ability to have a dedicated environment per branch for CI/CD. After necessary testing, features from these branches will merge back to master which will point to production environment. This allows better control over developing features without losing the CI/CD during development.

## Expected Behaviour

There should be ability to specify which OpenFaaS environment is listening for GitHub events from which branch and trigger CD accordingly.

## Current Behaviour

Currently only merges from master trigger CD.

alexellis:
One way this could be implemented at a low cost would be to have separate environments which are configured to only build from a certain named branch. Right now that named branch is `master`, but if we made this a configuration element you could define in the ofc-bootstrap tool then you could provision for instance two environments in around 200 seconds and then get the experience desired.

The `staging` environment would point at the `staging` branch and have a different set of URLs from the `master` environment (or production).

## GitHub validation

Code/check:

https://github.com/openfaas/openfaas-cloud/blob/ad4bb62902af4e597b60914cb03337f03ea5c45f/github-push/handler.go#L69

Test:

https://github.com/openfaas/openfaas-cloud/blob/4899ff2a3d10cd814a160e69e2c305172792e935/github-push/handler_test.go#L30

## stack.yml check

https://github.com/openfaas/openfaas-cloud/blob/7a72c8af543276ff2073d6bb08eefa1e8d87eb1e/git-tar/function/handler.go#L338

## Image reference format

https://github.com/openfaas/openfaas-cloud/blob/ad3fe9abfb1c40c15028034d0ae12599754cb1c0/git-tar/function/ops.go#L183

## UI for formatting GitHub link/URL to branch

https://github.com/openfaas/openfaas-cloud/blob/86852812d678f71b6f1f42b6f3e712a8d5ff585f/dashboard/client/src/components/FunctionTableItem/FunctionTableItem.jsx#L16

All of the above was found from a GitHub search for the text `"master"`. If we made all of this a config item i.e. `build_branch: dev/staging/master/` then it would be relatively straight-forward to set up multiple environments as per the request.

Alex

alexellis:
@martindekov can you confirm that GitLab is only limited to building from the master branch?

martindekov:
Let me check It out will comment with results

martindekov:
```
status: 200 response: refusing to build non-master branch: refs/heads/whatever
```
[...]

PR #438, opened 2019-04-21
Add building from different branches GitLab

martindekov:
Adding functionality which will ebable building functions
from multiple branches for gitlab-event function by adding
branches environmental variable with comma separated list

Signed-off-by: Martin Dekov <mdekov@vmware.com>

## Description

In Issue #425 me and @matipan agreed to split tasks, he did the majority of the Issue covering git-tar along with GitHub and dashboard this part covers GitLab only so I won't be closing the Issue with this PR.

Conceptual this covers gitlab-push only and you set `branches` environmental variable with comma separated list like so: `master,dev,staging` to enable which branches you would like to be able to build functions from. If unset, defaults to `master` branch only

## How Has This Been Tested?

Unit testing and manual tests here environmental `branches` set to `newbranch`:

### Building master branch fails:
image
### Building newbranch succeeds:
image

In dashboard function is displayed even when build from non-`master` branch.

They will be able to build functions from different branches depending on their needs.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/gitlab.yml
+++ b/gitlab.yml
@@ -24,7 +24,7 @@ functions:
     secrets:
       - gitlab-webhook-secret
       - payload-secret
-      - gitlab-api-token 
+      - gitlab-api-token
 
   gitlab-status:
     lang: go
@@ -41,8 +41,8 @@ functions:
       - gateway_config.yml
     secrets:
       - gitlab-api-token
-      - payload-secret 
-      
+      - payload-secret
+
   gitlab-push:
     lang: go
     handler: ./gitlab-push
@@ -54,8 +54,8 @@ functions:
     environment:
       write_debug: true
       read_debug: true
+      build_branch: "master"
     environment_file:
       - gateway_config.yml
     secrets:
       - payload-secret
- 
--- a/gitlab-push/handler.go
+++ b/gitlab-push/handler.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/alexellis/hmac"
 	"github.com/openfaas/openfaas-cloud/sdk"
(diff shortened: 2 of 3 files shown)


matipan:
I was under the impression that we were going to support only one branch per environment:
> One way this could be implemented at a low cost would be to have separate environments which are configured to only build from a certain named branch

The git-tar image that I pushed only builds from a single branch specified in an environment variable called `build_branch`. 

Should we support multiple branches or stick to one? @alexellis @martindekov

alexellis:
One branch per deployment for the initial configuration.

martindekov:
Yep, I will reverse that 👍

alexellis:
@martindekov how ready is this PR for review or merge?

gitlab.yml
```diff
@@ -41,21 +41,21 @@ functions:
...
-    image: functions/gitlab-push:0.2.1
+    image: functions/gitlab-push:0.3.0
     labels:
       openfaas-cloud: "1"
       role: openfaas-system
       com.openfaas.scale.zero: false
     environment:
       write_debug: true
       read_debug: true
+      branches: "master,"
```
alexellis: Can this still work even as `branches: "master"` without the `"master,`" (comma)?

martindekov: Yes I have added case in the unit tests but I will remove this functionality

martindekov:
Generally need to remove the multiple branches which will be easy I will do that by EOD otherwise it is tested e2e added unit tests and everything

alexellis:
Thanks I'll await that change :+1:

gitlab.yml
martindekov: Now this is removed 👍

alexellis:
Thank you Martin, and @matipan.

PR #439, opened 2019-04-21
Add improved message when template is missing

### LINKED ISSUE #417: Improve error message when function template isn't supported
## Expected Behaviour
When a template specified in a function's `stack.yml` doesn't exist within the community cluster the error message should clearly indicate that this is the cause of the failure.

## Current Behaviour

The current status API message is:
```
Deploy to OpenFaaS
tar(s) creation failed, error : open /tmp/rgee0/faas-demo/build/helloworld/config: no such file or directory
```
The list of supported templates is available [here](https://openfaas.o6s.io/signup#templates)

## Possible Solution
Determine possible causes for makeTar() to fail and try to determine cases where the template isnt supported.

https://github.com/openfaas/openfaas-cloud/blob/6c026ed11d20bc7a4703788e0dd6073e3de3c8d9/git-tar/function/ops.go#L110-L114

1. Create a non-supported function (python3-flask)
2. Push to an OFC enabled repo
3. Observe the status message

## Context
A user reported the issue in Slack.  Further investigation determined the cause, during the course of the investigation it was observed that the error message could be more informative.

## Your Environment
* FaaS-CLI version ( Full output from: `faas-cli version` ):
0.8.5

N/A

N/A

MacOS

https://github.com/rgee0/faas-demo

martindekov:
Adding message which signals that actually the
failure is due to missing template rather than the old
confusing output

Signed-off-by: Martin Dekov <mdekov@vmware.com>

## Description

Closes #417 

## How Has This Been Tested?

Manually e2e the image containing the change is `martindekov/of-git-tar:0.12.0` in order to test it yourself instead of checking out branches etc. and the status looks like this:
image

Users will now be able to see the actual error of their language not supported/missing/unconfigured instead of directly getting the failed to build error which was confusing

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/git-tar/function/ops.go
+++ b/git-tar/function/ops.go
@@ -647,3 +647,38 @@ func GitHubCloneURL(pushEvent sdk.PushEvent) (string, error) {
 
 	return cloneURL, nil
 }
+
+func checkCompatibleTemplates(stack *stack.Services, filePath string) (templatesErr error) {
+	templates, ioErr := existingTemplates(filePath)
+	if templatesErr != nil {
+		templatesErr = ioErr
+	}
+	for functionName, function := range stack.Functions {
+		for templateIndex, template := range templates {
+			if template == function.Language {
+				break
+			} else if templateIndex == len(templates)-1 && template != function.Language {
+				templatesErr = fmt.Errorf("Not supported language: `%s` for function: `%s`",
+					function.Language,
+					functionName,
+				)
+			}
+		}
+	}
+	return templatesErr
+}
+
+func existingTemplates(filePath string) ([]string, error) {
+	var existingTemplates []string
+	templatePath := fmt.Sprintf("%s/template", filePath)
+	files, ioErr := ioutil.ReadDir(templatePath)
+	if ioErr != nil {
+		return nil, fmt.Errorf("error while reading tempates directory: %s", ioErr)
+	}
+	for _, templateFolder := range files {
+		if templateFolder.IsDir() {
+			existingTemplates = append(existingTemplates, templateFolder.Name())
+		}
+	}
+	return existingTemplates, nil
+}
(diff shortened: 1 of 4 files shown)


stack.yml
```diff
@@ -47,7 +47,7 @@ functions:
   git-tar:
     lang: dockerfile
     handler: ./git-tar
-    image: functions/of-git-tar:0.11.2
+    image: functions/of-git-tar:0.12.0
```
alexellis: Can we sync ofc-bootstrap?

alexellis: Actually changing this file with a new set of code to be merged is hard because of timing. I have two push a new image immediately after merging your PR and if someone pulls in that window it may break for them.

It's harder for you but probably we should do code and config changes separately.

git-tar/function/ops.go
```diff
@@ -647,3 +647,38 @@ func GitHubCloneURL(pushEvent sdk.PushEvent) (string, error) {
 
 	return cloneURL, nil
 }
+
+func checkCompatibleTemplates(stack *stack.Services, filePath string) (templatesErr error) {
```
alexellis: Please can you add a test for both methods?

martindekov: Yes I will 👍

stack.yml
martindekov: Sounds reasonable, there is no problem 👍

git-tar/function/ops.go
martindekov: Done ✔️

stack.yml
martindekov: Reverted that 👍

git-tar/function/ops.go
martindekov: Added 👍

stack.yml
martindekov: Added in different PR

PR #441, opened 2019-04-24
Support custom build branches via env variable

matipan:

## Description
In order to provide a more CI/CD like experience for the users we decided to add an environment variable called `build_branch` that will have the branch that the users want to use for building their functions.
This PR changes the following functions:
* `buildshiprun`: I added a label for the deployment of each function: `com.openfaas.cloud.git-branch`. This label will have the branch that was used to deploy that function.
* `of-cloud-dashboard`: Add the git-branch to the response when calling `list-functions`. We then use that branch to show the link for the projects commits in the functions table.
* `git-tar`: I changed the format of the image name to include the branch that is being built. An example image name with this change: `matipan/matipan-faas-demo-slack-faces-result:latest-develop-cde0fb9`
* `github-push`: we used to validate that the branch trying to be built was only master. I changed it to check that the branch trying to be built is the one specified in `build_branch`. If `build_branch` is empty then we default to "master"

**Possible improvement**: add a `BuildBranch` function to the SDK package so that we don't have to replicate the same `buildBranch` func on each handler.

## How Has This Been Tested?

To test this I deployed the following functions on my own ofc installation:

* `matipan/buildshiprun:buildBranch`
* `matipan/of-git-tar:buildBranch`
* `matipan/github-push:buildBranch`
* `matipan/system-dashboard:buildBranch`

And set the `build_branch` to be `develop` for all deployments. [Here](https://github.com/matipan/faas-demo/commits/develop) you can see that ofc successfully built and deployed the functions using the develop branch.

The changes made to the dashboard and the image name can be seen here:
dashboard
imageName

And the new label that was added can be seen calling list functions:
newLabel

They are not since this default to using `master` whenever the git-branch label or the build_branch env var is not set.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/buildshiprun/handler.go
+++ b/buildshiprun/handler.go
@@ -222,6 +222,7 @@ func Handle(req []byte) string {
 				sdk.FunctionLabelPrefix + "git-sha":        event.SHA,
 				sdk.FunctionLabelPrefix + "git-private":    fmt.Sprintf("%d", private),
 				sdk.FunctionLabelPrefix + "git-scm":        event.SCM,
+				sdk.FunctionLabelPrefix + "git-branch":     buildBranch(),
 				"faas_function":                            serviceValue,
 				"app":                                      serviceValue,
 				"com.openfaas.scale.min":    scalingMinLimit,
@@ -694,4 +695,12 @@ func reportGitLabStatus(status *sdk.Status) {
 		log.Printf("unexpected error while reading response body: %s", bodyErr.Error())
 	}
 	status.CommitStatuses = make(map[string]sdk.CommitStatus)
+}
+
+func buildBranch() string {
+	branch := os.Getenv("build_branch")
+	if branch == "" {
+		return "master"
+	}
+	return branch
 }
--- a/dashboard/overview/handler.go
+++ b/dashboard/overview/handler.go
@@ -14,6 +14,7 @@ type UserData struct {
 	PublicURL      string
 	PrettyURL      string
 	QueryPrettyURL string
+	BuildBranch    string
 }
 
 // Handle a serverless request
@@ -37,6 +38,7 @@ func Handle(req []byte) string {
 
 	userData1.User = user
 	userData1.SelectedRepo = repo
+	userData1.BuildBranch = buildBranch()
 
 	log.Println("User: ", user, " Repo: ", repo)
 
@@ -49,4 +51,12 @@ func Handle(req []byte) string {
 	}
 
 	return tpl.String()
+}
+
+func buildBranch() string {
+	branch := os.Getenv("build_branch")
+	if branch == "" {
+		return "master"
+	}
+	return branch
 }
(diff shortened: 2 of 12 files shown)


alexellis:
Thank you for this PR and the detailed write-up and testing, this is brilliant. I'll ping someone to have a look over the code and give feedback if needed.

buildshiprun/handler.go
```diff
@@ -695,3 +696,11 @@ func reportGitLabStatus(status *sdk.Status) {
 	}
 	status.CommitStatuses = make(map[string]sdk.CommitStatus)
 }
+
+func buildBranch() string {
+	branch := os.Getenv("build_branch")
```
martindekov: We can use os.Lookupenv to distinguish between empty environmental variable and unset one

dashboard/overview/handler.go
```diff
@@ -50,3 +52,11 @@ func Handle(req []byte) string {
 
 	return tpl.String()
 }
+
+func buildBranch() string {
+	branch := os.Getenv("build_branch")
```
martindekov: Here also

git-tar/function/handler.go
```diff
@@ -360,3 +360,11 @@ func hasDockerfileFunction(functions map[string]stack.Function) bool {
 	}
 	return false
 }
+
+func buildBranch() string {
+	branch := os.Getenv("build_branch")
```
martindekov: We can also use lookupenv istead of getenv, also as I see this function appearing in some places, we can move `build_branc` in constant and the function itself in the SDK

github-push/handler.go
```diff
@@ -167,3 +167,11 @@ func reportGitHubStatus(status *sdk.Status) {
 		log.Printf("failed to report status, error: %s", reportErr.Error())
 	}
 }
+
+func buildBranch() string {
+	branch := os.Getenv("build_branch")
```
martindekov: We can also trim spaces just in case

github-push/handler_test.go
```diff
@@ -27,7 +27,7 @@ func Test_Handle_Push_InvalidBranch(t *testing.T) {
 		`{"ref":"refs/heads/staging"}`,
 	))
 
-	want := "refusing to build non-master branch: refs/heads/staging"
+	want := "refusing to build branch: refs/heads/staging"
```
alexellis: How about: `refusing to build target branch: x, want branch: y`

matipan:
> Hey there @matipan Great work! Minor comments on the way we extract the environmental variable, I believe the function and the naming can be exported in the SDK so I can reuse them in the GitLab PR. WDYT? I will also copy your message style smile

If definitely can! I added it as a possible improvement because we would new to make a different PR for that change, then create a release of openfaas-cloud with that change and finally change the `Gopkg.toml` of the functions. So that change would need to be on a different PR

github-push/handler_test.go
matipan: Good idea. Fixed :+1:

buildshiprun/handler.go
matipan: Do you need that distinction in this case? If the variable is unset or if it does not exist we need to return `master`, so if we use the LookupEnv we'll need to add another check for when the variable is empty

alexellis: No.. unset means master

alexellis:
CI passed, I'll merge :+1:

PR #455, opened 2019-05-15
Add network policies to limit ingress traffic

matipan:
This affects the namespaces `openfaas` and `openfaas-fn`.

## Description
In order to add rules to who can access the pods on the `openfaas` and `openfaas-fn` namespaces the network policies were modified. The rules are as follows:
* Pods in the `openfaas-fn` namespace only accept traffic from namespaces and pods that have the label `role: openfaas-system`
* Pods in the `openfaas` namespace only accept traffic from all pods in namespaces with the label `role: openfaas-system`, pods that have the label `role: openfaas-system` in the `openfaas-fn` namespace and finally pods from any namespace that have the label `app: nginx-ingress`(this is to allow traffic from the nginx ingress controller).

## How Has This Been Tested?

Deployed the network policies in my cluster and tried the following:
* Curl from a namespace that has no `role: openfaas-system` label to pods from `openfaas` and `openfaas-fn`, the connection timed-out since it never reached to the destination,
* Curl from the `openfaas` namespace to the `openfaas-fn` namespace and get a successful result,
* Curl from the `openfaas` namespace to other pods in the `openfaas` namespace and get a successful result,
* Curl from the `openfaas-fn` namespace to pods in the same namespace and get a successful result,
* Curl from the `openfaas-fn` namespace to pods in the `openfaas` namespace and get a successful result
* Call a function using the external domain and get a successful result(this means the network policy for the ingress is working)

They are affected only if they are using the network policies. If the are and they re-apply the policies everything should continue to work.
If they are not using the nginx ingress controller then the steps to follow are the same as they used to be.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -226,14 +226,12 @@ Create of-builder, of-buildkit:
 kubectl apply -f ./yaml/core
 ```
 
-(Optional) Deploy NetworkPolicy
+(Optional) Deploy NetworkPolicy. These policies set the following rules:
+* Pods in the `openfaas-fn` namespace only accept traffic from namespaces and pods that have the label `role: openfaas-system`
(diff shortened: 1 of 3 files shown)


alexellis:
@matipan please could you vendor this into ofc-bootstrap as a new YAML configuration option and have it installed by default from this point on?

PR #456, opened 2019-05-16
Ytel Testing

dpayne-ytel:
## OpenFaas Cloud Access
I have been using OpenFaas in production for the last couple of months. Would like to try out the cloud platform

### DIFF
--- a/CUSTOMERS
+++ b/CUSTOMERS
@@ -88,3 +88,4 @@ xmudrii
 zeerorg
 zhenfeng-zhu
 paurosello
+dpayne-ytel


rgee0:
All commits need to be signed off.  You probably only need one commit for this, so perhaps try squashing the commit that isn't signed into the commit that is signed and then force push.

alexellis:
What Richard said.. also you might find it easier just to start a new PR?

dpayne-ytel:
I removed the erroneous commit. The one remaining is signed off

PR #526, opened 2019-10-16
Signed-off-by: Morenko Roman <rmorenko@neoflex.ru>

romanmorenko:
I learned about OpenFaaS from the Micronaut documentation https://micronaut.io/documentation.html. I want to try a platform to consider using in our new projects. I want to look at the development process and deployment using simple applications as an example.


alexellis:
Access is granted, for future reference the portal for the community cluster is: https://github.com/openfaas/community-cluster

PR #550, opened 2019-11-13
Add helm chart for installing edge-auth and of-builder

### LINKED ISSUE #532: [Feature] Helm chart for of-builder and edge-router/auth
## Task

We should add a helm chart for everything in the [yaml folder](https://github.com/openfaas/openfaas-cloud/tree/master/yaml) so that these components can be templated easily.

* core - edge-router/auth (see values set in ofc-bootstrap - https://github.com/openfaas-incubator/ofc-bootstrap/tree/master/templates)
* core - of-builder (see values set in ofc-bootstrap - https://github.com/openfaas-incubator/ofc-bootstrap/tree/master/templates)
* network-policy - network policies (on/off)

This is now handled by ofc-bootstrap [1], but it means maintaining two copies of the files - templated and static.

The helm chart can eventually replace the static YAML files, but both will be needed for compatibility in the short-term.

Start with the scaffolded example:

```
mkdir charts
cd charts
helm create
```

Then add in each resource making sure to template the name, namespace, tolerations and etc.

Once completed, ofc-bootstrap should be changed to use the helm chart and to stop using its custom copies of the files.

[1] https://github.com/openfaas-incubator/ofc-bootstrap

Waterdrips:
## Description
This adds a Helm chart for some of the OFC components, this will allow easier installation for ofc (directly of via ofc-bootstrap)

The idea is to use these to replace some static/templated yaml in ofc & ofc-bootstrap eventually.

We need to add some automated testing and stuff to this, but i couldnt get `ct`(https://github.com/helm/chart-testing) setup because it strugles when there isnt an existing helm repo setup to "diff" against...

@alexellis - this will need serving as a helm repo somewhere - and installation instructions etc need amending? 

closes #532 

## How Has This Been Tested?

Manually removing my edge-auth, edge-router and of-builder in by github-ofc lined cluster, then installing these deployments & services using the helm chart

example output here: https://gist.github.com/Waterdrips/81b530784e8731d930d3ed52e21ac65b 
(with mock values)

this is the minimum number of things set required:
```sh
helm install --namespace openfaas openfaas-cloud/openfaas-cloud \
  --set edgeAuth.clientId=my-client-id  \
  --set edgeAuth.clientSecret=my-client-secret \
  --set global.rootDomain=example.com \
  --set global.customersURL=https://my-customers-url.CUSTOMERS 
```
People can also enable ECR or Gitlab by setting: 
```sh
  --set global.enableECR=true 
 ```
and 
```sh
-- set OAuthProvider=gitlab
--set edgeAuth.OAuthProviderBaseURL=some.gitlab.example.com 
```
This is a new installation method for these components and would not be mandatory. 

## Checklist:

I have:

- [ ] added unit tests


alexellis:
`  --set edgeAuth.clientSecret=some-secret` - this should perhaps be the name of a secret?

charts/openfaas-cloud/templates/edge-auth-dep.yaml
```diff
@@ -0,0 +1,81 @@
...
+      app: {{ .Values.edgeAuth.name }}
+  template:
+    metadata:
+      annotations:
+        prometheus.io.scrape: {{ .Values.global.prometheusScrape | quote }}
+      labels:
+        app: {{ .Values.edgeAuth.name }}
+    spec:
+      volumes:
+        - name: jwt-private-key
```
alexellis: Do we ever template these values or names? (if not, that's fine)

Waterdrips:
> ` --set edgeAuth.clientSecret=some-secret` - this should perhaps be the name of a secret?

the "some-secret" bit was me redacting my secret key, it would be something like 9ew8ywihbf8w6fg8w74gf8w 
Is that what you mean?

charts/openfaas-cloud/templates/of-builder-dep.yaml
```diff
@@ -0,0 +1,79 @@
...
+{{- end }}
+        - name: of-buildkit
+          args: ["--addr", "tcp://0.0.0.0:1234"]
+          image: {{ .Values.ofBuilder.buildKitImage }}:{{ .Values.ofBuilder.buildKitTag }}
+          imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+          ports:
+            - containerPort: 1234
+              protocol: TCP
+          securityContext:
+            privileged: true
```
alexellis: There is a way to run as `privileged=false` and non-root which it would be good to have as an option

> Using Rootless mode (*.rootless.yaml) is recommended because Rootless mode image is executed as non-root user (UID 1000) and doesn't need securityContext.privileged.

https://github.com/moby/buildkit/tree/master/examples/kubernetes

alexellis: It might be something to look at in this PR or with a follow-up issue and PR, this chart will certainly make it easier to experiment.

charts/openfaas-cloud/templates/edge-auth-dep.yaml
Waterdrips: this is from the README for the edge-auth component

```sh
kubectl -n openfaas create secret generic jwt-private-key --from-file=./key
kubectl -n openfaas create secret generic jwt-public-key --from-file=./key.pub
```

then we use these named secrets, ofc-bootstrap also uses these names, so I think for now we dont need a template variable.

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
+# Default values for openfaas-cloud.
+# This is a YAML-formatted file.
+
+## ofBuilder is the section for of-builder configuration specific settings
+ofBuilder:
+
+  ## names is the name of the deployment for the of-builder component
+  name: of-builder
```
alexellis: I've not seen this done before, or perhaps we keep the name then prefix with the release name?

https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/templates/queueworker-dep.yaml#L11

I wonder if it'd be simpler if the name wasn't templated?

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+  ## image is the container image that is used for the deployment of of-builder
+  image: openfaas/of-builder
```
alexellis: We have this in one field with faas-netes to help with use with CD tools like Flux and Argo - https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/values.yaml#L24

It's easier for them to work with.

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+  image: openfaas/edge-auth
+
+  ## tag is the image tag to be deployed when deploying this component
+  tag: 0.6.2
+
+  ## replicaCount is the number of pod replicas for the the deployment
+  replicaCount: 1
+
+  ## OAuthProvider is either github or gitlab depending on who is providing OAuth for your cluster
+  OAuthProvider: github
```
alexellis: Should it be capitalised or `oauthProvider`?

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+  ## OAuthProviderBaseURL can be left blank for github, Gitlab users need to provide the url for their gitlab instance
+  OAuthProviderBaseURL:
+
+  ## clientId is generated by your Oauth App, and needs to be set here to connect correctly
+  clientId:
+
+  ## writeDebug is a setting to tell edge-auth to write debug messages
+  writeDebug: false
+
+  ## clientSecret is the value from yur OAuth application
```
alexellis: `your`

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+  container: openfaas/edge-router
+
+  ## tag is the tag for edge-router deployment
+  tag: 0.7.4
+
+## global values that are re-used across multiple components
+global:
+
+  ## rootDomain is the root domain of the ofc installation
+  rootDomain:
```
alexellis: Would it be a good idea to have example.com, or not?

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+  enableECR: false
+
+  ## imagePullPolicy is the setting sent to components to state if they should always pull a new image or use a cached version if available
+  imagePullPolicy: Always
+
+  ## enableOAtuth2 is a setting to enable oauth2 or not.
+  enableOAuth2: true
+
+  ## openfaasNamespace this is used where we need to communicate with core openfaas components
+  openfaasNamespace: openfaas
```
alexellis: 👏 this will be important

charts/openfaas-cloud/templates/of-builder-dep.yaml
Waterdrips: Its easy to add that in now, so ill just pop it in. 👍

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+  ## openfaasFnNamespace this is where the default openfaas-fns live, like "echo" so our core services can call them
+  openfaasFnNamespace: openfaas-fn
```
alexellis: `openfaasFunctionsNamespace`?

charts/openfaas-cloud/templates/edge-router-dep.yaml
```diff
@@ -0,0 +1,50 @@
...
+      annotations:
+        prometheus.io.scrape: {{ .Values.global.prometheusScrape | quote }}
+      labels:
+        app: {{ .Values.edgeRouter.name }}
+    spec:
+      containers:
+        - name: {{ .Values.edgeRouter.name }}
+          image: {{ .Values.edgeRouter.container }}:{{ .Values.edgeRouter.tag }}
+          imagePullPolicy: Always
+          livenessProbe:
```
alexellis: As a note - eventually the `http` checks will cause issues for Istio users. I know the YAMLs don't cover this yet, but in the future we could add exec/http selection via the approach in faas-netes https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/values.yaml#L7

alexellis:
@csakshaug this will affect ofc-bootstrap so that you're aware. In the long-run it will be a net gain and make maintenance easier.

charts/openfaas-cloud/templates/edge-router-dep.yaml
```diff
@@ -0,0 +1,50 @@
...
+    metadata:
+      annotations:
+        prometheus.io.scrape: {{ .Values.global.prometheusScrape | quote }}
+      labels:
+        app: {{ .Values.edgeRouter.name }}
+    spec:
+      containers:
+        - name: {{ .Values.edgeRouter.name }}
+          image: {{ .Values.edgeRouter.container }}:{{ .Values.edgeRouter.tag }}
+          imagePullPolicy: Always
```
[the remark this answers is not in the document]
Waterdrips: Ah that should be templated. Good spot

charts/openfaas-cloud/values.yaml
Waterdrips: it wont work either way. We could have it in there as an ... example ... or we could add that info to the comment.

I am trying to work out a way to used the `required` functionality, but I cant get it to work with the printf functionality. That way, if its not set by the user the deployment fails.

charts/openfaas-cloud/values.yaml
Waterdrips: Yeah i totally misread the "best practices" doc, change incoming 🚀

csakshaug:
> @csakshaug this will affect ofc-bootstrap so that you're aware. In the long-run it will be a net gain and make maintenance easier.

I agree

charts/openfaas-cloud/templates/edge-router-dep.yaml
Waterdrips: I have added the same functionality as faas-netes's chart, with an option to switch to an exec check

charts/openfaas-cloud/templates/tls/auth-domain-cert.yml
```diff
@@ -0,0 +1,13 @@
+---
```
alexellis: I forgot to say that "ingress shim" can create the certs, and we can do without the `Certificate` object going forward.

alexellis: (Like we do in k3sup)

charts/openfaas-cloud/templates/tls/issuer-staging.yml
```diff
@@ -0,0 +1,32 @@
+apiVersion: cert-manager.io/v1alpha2
```
alexellis: Now I see why you were trying to add helm tests ;-)

charts/openfaas-cloud/templates/ingress-wildcard.yml
```diff
@@ -0,0 +1,33 @@
...
+  {{ end }}
+  rules:
+  - host: {{ printf "*.%s" .Values.global.rootDomain | quote }}
+    http:
+      paths:
+      - path: /
+        backend:
+          serviceName: edge-router
+          servicePort: 8080
+  - host: {{ printf "gateway.%s" .Values.global.rootDomain | quote }}
```
alexellis: Please can we drop gateway both here and in orc-bootstrap? It isn't required and could expose some functions we want to keep private, Linkerd also doesn't like it when doing the host mapping.

alexellis: That'll leave us with -> IngressController -> Edge Router -> GW

Waterdrips:
Im having a bit of trouble with the sealed-secrets stuff - think its RBAC. Working on it but its slow going !

alexellis:
> Im having a bit of trouble with the sealed-secrets stuff

Can you explain what you mean? Are you adding SealedSecrets via this PR, or having issues in your local setup? Let me know if I can help.

Waterdrips:
> > Im having a bit of trouble with the sealed-secrets stuff
> 
> Can you explain what you mean? Are you adding SealedSecrets via this PR, or having issues in your local setup? Let me know if I can help.

When i installed the components on this chart (after deleting them via kubectl) the sealed-secrets component started sending lots of error events to `kubectl get events -A` i then reinstalled it and it seemed to work again.

Waterdrips:
I need to add Cloudflare DNS01 support to this because this was added in this PR https://github.com/openfaas-incubator/ofc-bootstrap/pull/160 to ofc-bootstrap and will be required downstream

charts/test/ingress_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"io/ioutil"
+	"reflect"
+	"testing"
+
+	"gopkg.in/yaml.v2"
+
+)
+
+func Test_(t *testing.T) {
```
alexellis: Would be better with a name?

Waterdrips: DOH! defiantly.

charts/test/ingress_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	annotations := make(map[string]string)
+	annotations["kubernetes.io/ingress/class"] = "nginx"
+	annotations["nginx.ingress.kubernetes.io/limit-connections"] = "20"
+	annotations["nginx.ingress.kubernetes.io/limit-rpm"] = "600"
+
+
+	want := ingressAuth{
+		ApiVersion: "appa/v1",
```
alexellis: appa -> apps?

alexellis: I like the approach, but is it working? Ingress doesn't have apps/v1.

charts/test/ingress_test.go
```diff
@@ -0,0 +1,52 @@
...
+		Metadata:   MetadataItems{
+			Name:        "openfaas-auth-ingress",
+			Annotations: annotations,
+		},
+		Spec:       Spec{},
+	}
+	_, _ = helmRunner(parts...)
+
+
+	data, err := ioutil.ReadFile("./tmp/openfaas-cloud/templates/ingress/ingress-auth.yml")
```
alexellis: Check err before trying to use data.

charts/test/ingress_test.go
```diff
@@ -0,0 +1,52 @@
...
+	want := ingressAuth{
+		ApiVersion: "appa/v1",
+		Kind:       "Ingress",
+		Metadata:   MetadataItems{
+			Name:        "openfaas-auth-ingress",
+			Annotations: annotations,
+		},
+		Spec:       Spec{},
+	}
+	_, _ = helmRunner(parts...)
```
alexellis: Doesn't this return anything that could be checked? The output path for instance would be better returned than hard-coded if possible. Can it throw an error like helm not found?

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,125 @@
...
+# This is a YAML-formatted file.
+
+## ofBuilder is the section for of-builder configuration specific settings
+ofBuilder:
+
+  ## image is the container image and tag that is used for the deployment of of-builder
+  image: openfaas/of-builder:0.7.2
+
+  ## buildKitImage is the buildkit container image that is used within the deployment for building containers
+  buildKitImage: moby/buildkit
```
alexellis:
```yaml
buildkit:
  image: moby/buildkit:v0.6.2
  replicas: 1
  privileged: true
```

https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/values.yaml#L24

charts/openfaas-cloud/templates/ingress/ingress-auth.yml
```diff
@@ -0,0 +1,26 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: openfaas-auth-ingress
+  namespace: {{ .Values.global.coreNamespace }}
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
```
alexellis: Just a thought, would it make sense to template the ingress class? If not now, in a future PR/issue?

charts/openfaas-cloud/templates/ingress/ingress-auth.yml
```diff
@@ -0,0 +1,26 @@
...
+    nginx.ingress.kubernetes.io/limit-connections: "20"
```
alexellis: These limits would be good to template since it may not be high enough for corporate users like DX cc @csakshaug

charts/openfaas-cloud/templates/ingress/ingress-wildcard.yml
```diff
@@ -0,0 +1,26 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: openfaas-ingress
+  namespace: {{ .Values.global.coreNamespace }}
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+    nginx.ingress.kubernetes.io/limit-connections: "20"
```
alexellis: Could also be templated.

charts/openfaas-cloud/templates/ingress/ingress-wildcard.yml
```diff
@@ -0,0 +1,26 @@
...
+    nginx.ingress.kubernetes.io/limit-rpm: "600"
+  labels:
+    app: faas-netesd
+spec:
+  {{- if .Values.tls.enabled }}
+  tls:
+  - hosts:
+    - {{ printf "*.%s" .Values.global.rootDomain | quote }}
```
alexellis: `printf` is a new one on me in helm charts, what's the advantage?

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+  replicaCount: 1
+
+buildKit:
+  image: moby/buildkit:v0.6.2
+  privileged: true
+
+
+edgeAuth:
+  image: openfaas/edge-auth:0.6.3
+  replicaCount: 1
```
alexellis: `replicas: 1` please

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+  enabled: false
+  ## email is used to send cert-renewal emails, it is required by LetsEncrypt
+  email:
+
+  ## Default to using the staging issuer, there's a limit to 5 certs per domain on the prod issuer, so only when you are
+  ##  happy should you swap this to prod
+  issuerType: staging
+
+  ## dnsService, one of "route53", "digitalocean" , "cloudflare" or "clouddns"
+  dnsService:
```
alexellis: How do you "select" a provider?

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+    email:
+
+  clouddns:
+    projectID:
+
+## global values that are re-used across multiple components
+global:
+
+  ## rootDomain is the root domain of the ofc installation o6s.io for example
+  rootDomain:
```
alexellis: Would be good to have an actual sample value here `of.example.com`

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+  prometheusScrape: false
```
alexellis: Not sure we'd want to template `prometheusScrape`? We used `prometheusScrape` to turn off Prometheus from scraping services which have no metrics

charts/openfaas-cloud/values.yaml
```diff
@@ -0,0 +1,73 @@
...
+  ## customersURL is the url to the raw file that containers the list of users who are able to use this installation
+  customersURL:
+  ## scheme is the http/https setting for the installation
+  scheme: https
+  enableECR: false
+  imagePullPolicy: IfNotPresent
+  enableOAuth2: true
```
alexellis: `enableOAuth2 ` I think we have this in two places?

charts/openfaas-cloud/templates/ingress/ingress-wildcard.yml
Waterdrips: if you want to pass the result to `quote`

charts/openfaas-cloud/values.yaml
Waterdrips: I was leaving it empty to force users to set it, they would get a `global.rootDomain is required` type message if they didnt customise, otherwise they will happily install a chart which would never work?

charts/openfaas-cloud/values.yaml
Waterdrips: theres a section on configuring OAuth, it could be moved there?

charts/openfaas-cloud/values.yaml
Waterdrips: --set dnsService=route53

We could make an assumption, but that assumption assumes they have created the correct secrets ?

Waterdrips:
@alexellis I have addressed the templating comments. 

The comments about default variables, what do you want to do? We can add defaults to everything but that could lead to users not knowing what they have to specify at a minimum? I'm using a required check on things we don't have a default for, so the install will error with a message.

Waterdrips:
@alexellis comments addressed from earlier.

I think if we generally agree with the approach and implementation we could get this in and start to iterate on it?

alexellis:
Merging as discussed with yourself and @csakshaug on our call last week. We will iterate on it from here. 

Please add a `make charts` command as per the faas-netes repo into the Makefile.

alexellis:
We also need the README.md file to be populated with something, even if not extensive. There's a blank Makefile too.

PR #553, opened 2019-11-16
Fix issue in dashboard with non-auth deployments

Waterdrips:
## Description
A deployment that didn't have auth enabled would fail
to list functions, or pass any requests through to the
backend that were destined for the api as the new auth
put in expected a token, and didn't consider the no-auth
deployment solution.

## How Has This Been Tested?

deployed the new dashboard, it works when i had auth enabled.
Then i removed auth from my cluster by setting auth_url to echo fn.
The dashboard now lists functions correctly when theres no cookie.

Any users not using auth will need to deploy the new dashboard container to fix the issue.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/dashboard/of-cloud-dashboard/handler.js
+++ b/dashboard/of-cloud-dashboard/handler.js
@@ -54,7 +54,7 @@ module.exports = (event, context) => {
   if (/^\/api\/(list-functions|system-metrics|pipeline-log|function-logs).*/.test(path)) {
 
     // See if a user is trying to query functions they do not have permissions to view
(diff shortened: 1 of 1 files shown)


Waterdrips:
There's no cookie. As ingress comes via the edge-router, which validates auth, if we dont have a cookie then theres no auth enabled.

Waterdrips:
I much prefer the env var idea, I'll get this put in and updated

alexellis:
@Waterdrips please can you rebase this?

Waterdrips:
Fixes for this were incorporated into #556

PR #555, opened 2019-11-19
Rename system-metrics to metrics #551

### LINKED ISSUE #551: Rename system-metrics function to metrics
functions prefixed with system- would generally be expected to be world accessible, system-metrics is not world accessible and should therefore be renamed

## Expected Behaviour

no functional change

## Impact on existing deployments
An administrator would need to delete the function `system-metrics` and deploy the newly created `metrics` function using faas-cli.

Waterdrips:
## Description

The system- prefixed functions are generally exposed
to the ingress, so we have renamed system-metrics to
metrics so the naming convention is the same

closes #551 

## How Has This Been Tested?
Deployed the new metrics function (rename only) to my OFC installation, deleted the system-metrics deployment and service.

Checked that metrics are sowing on the system-dashboard, both on the functions table, and on the functions specifics page.

no functional change

Users will need to remove the system-metrics deployment and service, and deploy the new metrics function and system-dashboard.

**The stack.yaml file will need system-dashboard --> dashboard for the image name when a release is created.** 

## Checklist:

I have:

- [ ] added unit tests


stack.yml
```diff
@@ -206,10 +206,10 @@ functions:
       write_debug: true
       read_debug: true
 
-  system-metrics:
+  metrics:
     lang: go
-    handler: ./system-metrics
-    image: functions/system-metrics:0.1.1
+    handler: ./metrics
+    image: functions/metrics:0.1.1
```
alexellis: Can we retain the image name since the new name doesn't exist yet, so it may break when merged?

Waterdrips: ah yes - excellent point.

PR #361, opened 2018-12-11
Add user/org filtering to dashboard

ivanayov:
After listing functions both from the user and the organizations
they are part of, function owner should be available in the UI in
order to indicate if the function is owned by orgs or by user

Signed-off-by: Ivana Yovcheva <iyovcheva@vmware.com>

Added owner filtering to function table

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

Closes #359 

## How Has This Been Tested?

screen shot 2018-12-11 at 22 36 01

with `docwareiy/of-cloud-dashboard:0.3.4-show-owner`

Latest change: (the colors are from following PR by @bartsmykla based on this)
screen shot 2018-12-14 at 10 10 21

Dashboard: https://system.smykla.io/dashboard/bartsmykla

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required) N/A
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [ ] added unit tests N/A

PR #556, opened 2019-11-23
Show all Orgs that a user is a member of in user's dashboard

### LINKED ISSUE #347: [Feature] Show and filter functions from my orgs in my dashboard

Waterdrips:
## Description
We can now display all organisation's functions when a user is a public member and Oauth is enabled.
This shows as Org functions in a new column on dashboard with links pointing to that organisation's urls (a user will then be in that org's dashboard area, the breadcrumb link will be back to that org)

Lots of this code is from an old PR,
closes #347 
closes #361 

See screenshot:
Screenshot 2019-12-04 at 08 10 28

## How Has This Been Tested?

Deployed to my own OFC installation, all links checked from dashboard etc.

New deployment of the system-dashboard.
This is extra functionality and should improve usability of the OFC dashboard.

## Checklist:

I have:

- [ ] added unit tests


alexellis:
Thanks for working on this. Can I try it out somewhere?

Waterdrips:
@alexellis is authorised on my OFC installation: https://system.heyal.uk/dashboard/

I have just deployed from the tip of this branch, so it matches. 

If you want to install the dashboard on your own cluster you can use this image: 
`heyal/of-cloud-dashboard:demo-v1`

alexellis:
I'm seeing OF which also stands for OpenFaaS. This is a little confusing. I wonder if we need to rethink the original idea about shortening to initials? How is the colour calculated? On full name or initials? What if you add an org called OpenFields or similar?

Waterdrips:
I did look at grabbing GitHub avatars, but this wouldn't work for gitlab. 

In terms of colour and initials this is code that was pulled from the old PR (#361)
Colour functions point at this : https://stackoverflow.com/questions/3426404/create-a-hexadecimal-colour-based-on-a-string-with-javascript

Initials come from 
```js
const genOwnerInitials = (owner) => (
  (owner.split(/[-_]+/).length > 1) ?
    owner.split(/[-_]+/)[0].substring(0, 1).concat(owner.split(/[-_]+/)[1].substring(0, 1)) :
      owner.split(/[-_]+/)[0].substring(0, 2)

);
```
If the owner has a - or _, use the first letter of first 2 sections
If not, use first 2 letters

Might be better without this stuff, and just use the org/user name

Waterdrips:
hi @alexellis, iv rebased this pr and re-tested (With and without auth)

Have you thought any more about the initials vs full org/username in the table?

alexellis:
> Might be better without this stuff, and just use the org/user name

^ In the screenshot it would probably take the same amount of room, let's err on the side of simple and try that?

Waterdrips:
@alexellis - This is without formatting
Screenshot 2019-12-02 at 16 39 02

alexellis:
Can I try this? Do you have an environment up?

alexellis:
How does it look on mobile now for responsive web? Can you take a screenshot in the Chrome preview?

Waterdrips:
> @alexellis is authorised on my OFC installation: https://system.heyal.uk/dashboard/
> 
> I have just deployed from the tip of this branch, so it matches.

^ Its available there.


Responsive view of the table is not ideal, it scrolls sideways, this has not changed from previous versions.
Screenshot 2019-12-06 at 09 24 50
Screenshot 2019-12-06 at 09 24 56

alexellis:
What do we want to drop for this viewport?

alexellis:
May I suggest: invocations replicas sha?

Waterdrips:
How about this: 
Screenshot 2019-12-07 at 11 24 42

dashboard/client/src/components/FunctionTableItem/FunctionTableItem.jsx
```diff
@@ -20,7 +20,14 @@ const genRepoUrl = ({ gitRepoURL, gitBranch }) => {
   return `${gitRepoURL}/commits/${gitBranch}`
 };
 
-const FunctionTableItem = ({ onClick, fn, user }) => {
+const genOwnerInitials = (owner) => (
```
alexellis: Do we still need this function?

dashboard/of-cloud-dashboard/handler.js
```diff
@@ -207,7 +212,9 @@ var isResourceInTokenClaims = function (path, queryString, user, organisations)
   if (user === queryString["user"]) {
     return true
   }
-  return organisations.indexOf(queryString["user"]) >= 0
+
+  let orgs = organisations.split(",");
```
alexellis: Can `organisations` be undefined? What if it was zero lengthed?

Waterdrips: No it can't be undef, if we dont have a cookie we return above, otherwise its an empty string.

and "".split(",") --> [""], so no error

Waterdrips: No it can't be undef, if we dont have a cookie we return above, otherwise its an empty string.

and "".split(",") --> [""], so no error

dashboard/client/src/components/FunctionTableItem/FunctionTableItem.jsx
Waterdrips: no

PR #557, opened 2019-11-23
Replace hard coded github.com with variable

Waterdrips:
## Description

There was a link on the function detail page that was pinned to
github.com (as string) rather than using the function's reported
location, which could have been gitlab

## How Has This Been Tested?

Deployed to my OFC cluster, links render for github.com still (dont have a gitlab cluster to test on)

New Dashboard required, this should fix an issue with Gitlab link, which nobody has reported, but might happen

## Checklist:

I have:

- [ ] added unit tests


alexellis:
An idea for testing - why don't you edit the annotation or label and see if the link updates?

Waterdrips:
Will do

Waterdrips:
That change to the annotation changed the link as expected. Thanks for that idea Alex!

alexellis:
Remind me to regenerate and update the dashboard container for this change

PR #558, opened 2019-11-24
[WIP] Allow user to use tail and/or since param when querying logs

rajatjindal:
## Description

Allow user to use tail and/or since param when querying logs

## How Has This Been Tested?

Deployed the functions locally and tested

Tested the functions manually using API as well as the UI

N/A

## Checklist:

I have:

- [ ] added unit tests

I still need help to place the elements correctly on the UI.

### DIFF
--- a/dashboard/client/src/pages/FunctionLogPage.jsx
+++ b/dashboard/client/src/pages/FunctionLogPage.jsx
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import AceEditor from 'react-ace';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import {Card, CardHeader, CardBody, Button} from 'reactstrap';
+import {Card, CardHeader, CardBody, Button, Input, InputGroup, InputGroupText, Container, Col, Row} from 'reactstrap';
 
 import 'brace/mode/sh';
 import 'brace/theme/monokai';
@@ -25,18 +25,23 @@ export class FunctionLogPage extends Component {
       log: '',
       functionName,
       user,
-      fetchError: false
+      fetchError: false,
+      numLines: 100,
+      since: 0
     };
+
+    this.changeNumLines = this.changeNumLines.bind(this);
+    this.changeSince = this.changeSince.bind(this);
   }
 
   componentDidMount() {
-    const { functionName, user } = this.state;
+    const { functionName, user, numLines, since } = this.state;
 
     this.setState({ isLoading: true });
 
     const longFnName = `${user.toString().toLowerCase()}-${functionName}`
       try {
-          functionsApi.fetchFunctionLog({longFnName, user}).then(res => {
+          functionsApi.fetchFunctionLog({longFnName, user, numLines, since }).then(res => {
             this.setState({isLoading: false, log: res});
 
           })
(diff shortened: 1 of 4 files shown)


rajatjindal:
requesting feedback/help from @Waterdrips

dashboard/client/src/pages/FunctionLogPage.jsx
```diff
@@ -25,18 +25,23 @@ export class FunctionLogPage extends Component {
       log: '',
       functionName,
       user,
-      fetchError: false
+      fetchError: false,
+      numLines: 0,
```
Waterdrips: This should be a sensible default, say 100?

function-logs/handler.go
```diff
@@ -28,6 +32,16 @@ func Handle(req []byte) string {
 			if len(userQuery) > 0 {
 				user = userQuery
 			}
+			numLines = vals.Get("numLines")
+			sinceStr := vals.Get("since")
+			if sinceStr != "" {
+				var err error
+				since, err = strconv.Atoi(sinceStr)
```
Waterdrips: `since, err := strconv.Atoi(sinceStr)`
then you dont need to declare var err error.

function-logs/handler.go
```diff
@@ -28,6 +32,16 @@ func Handle(req []byte) string {
 			if len(userQuery) > 0 {
 				user = userQuery
 			}
+			numLines = vals.Get("numLines")
```
Waterdrips: this should be a `:=` too, then `vat numLines string` and `var since int` are not required.

dashboard/client/src/pages/FunctionLogPage.jsx
```diff
@@ -25,18 +25,23 @@ export class FunctionLogPage extends Component {
...
+      since: 1
```
Waterdrips: since default might want to be 0, so it fetches the numLines by default

function-logs/handler.go
rajatjindal: We need since outside this function, thats why i didnt do it using :=

dashboard/client/src/pages/FunctionLogPage.jsx
rajatjindal: Will make that change

dashboard/client/src/pages/FunctionLogPage.jsx
rajatjindal: Sure, will make that change

rajatjindal:
The ui looks ugly as hell with my changes, will attach screenshot as soon as i m near laptop

function-logs/handler.go
rajatjindal: We need these vars out of scope of that if condition, therefore its like that

rajatjindal:
Screen Shot 2019-11-25 at 8 39 53 PM
Screen Shot 2019-11-25 at 8 40 07 PM
Screen Shot 2019-11-25 at 8 40 18 PM

rajatjindal:
@Waterdrips I have made the changes requested and attached the screenshots. 

Like I said, it looks ugly, so I will need your help to fix that :)

rajatjindal:
between, it seems like from k8s perspective, since and tail are mutually exclusive? so if we provide tail, it ignores since.

should we disable that from UI to ensure only one of them is provided

Waterdrips:
I wonder if we should change the input to a dropdown, and have some pre-selected values, 100, 200, 500, 1000? or something. 

As for the `since` im not sure if we need it? does the logs backend accept a "give me everything" type payload? we could offer that up as `>1000 ` or `all`

rajatjindal:
requesting feedback from @alexellis @LucasRoesler

alexellis:
Whilst this does look nice, I think I mentioned this preference already - the OFC UI should be hard coded to a set value. We had 5 mins now perhaps it should be 100 lines?

rajatjindal:
Hi Alex, we can make that change, and remove overriding the values from the UI. However I am curious to know as to why we do not want to allow users to be able to say I want 100 lines or 500?

alexellis:
There are various reasons for this:

* the logs are pulled directly from the Kubernetes API which is heavy-weight and easy to abuse
* the logs are likely to be very limited due to the scale to zero behaviour - whenever scaling down the logs are removed
* this was not requested and is not in scope 

There should simply be one server-side change to faas-netes, then another minor change in the backend code which pulls the metrics to send a number of lines instead of a time period.

I'll close this PR as it's currently misleading, let's start a new one with the "surgical" change that was approved for OFC.

alexellis:
In future, please propose all changes before working on a PR, it will help avoid wasted time.

PR #321, opened 2018-10-30
Add validation for custom templates

ivanayov:
This adds url validation for custom templates and edits the test
to cover previous issues in function code.

Signed-off-by: Ivana Yovcheva <iyovcheva@vmware.com>

Closes #319 

## How Has This Been Tested?

Unit tested + repeated the tests from #264

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [x] added unit tests

PR #566, opened 2019-12-09
Add validation for custom templates

### LINKED ISSUE #319: Validate custom template URLs

Waterdrips:
This adds url validation for custom templates and edits the test
to cover previous issues in function code.

## Description
Rebased the initial PR by Ivana, I have changed it slightly so it doesn't die if it finds an invalid template, it instead logs it to git-tar output.

I decided that if anyone has any old templates in there cluster config we don't want to break their installations, just give some more gentle nudges if people check the logs.

closes #321
closes #319 
## How Has This Been Tested?

unit tests done by Ivana,

I have deployed this new service onto my cluster and put in an invalid template, this gives error output in the logs, but still builds if we are using a valid template

People will get new log lines if they have templates that are not valid.
Users should run a build and follow the logs to see if they ahve enabled any invalid templates:

`kubectl logs -n opoenfaas-fn deploy/git-tar -f`

A user will need to deploy this new service first with:
( we are assuming the next version of git-tar will be `0.16.1` if this is not the case then replace the version beelow`)

`kubectl set image deployment/git-tar git-tar=functions/git-tar:0.16.1 --record -n openfaas-fn`

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/git-tar/function/ops.go
+++ b/git-tar/function/ops.go
@@ -41,35 +41,48 @@ type cfg struct {
 	Frontend *string `json:"frontend,omitempty"`
 }
 
-func parseYAML(pushEvent sdk.PushEvent, filePath string) (*stack.Services, error) {
+func parseYAML(filePath string) (*stack.Services, error) {
 	envVarSubst := false
 	parsed, err := stack.ParseYAMLFile(path.Join(filePath, "stack.yml"), "", "", envVarSubst)
 	return parsed, err
 }
 
 func fetchTemplates(filePath string) error {
-	templateRepos, err := formatTemplateRepos()
-	if err != nil {
-		return fmt.Errorf("Failed to format template repos: %t", err)
+	templateRepos, errors := formatTemplateRepos()
+
+	if err := joinErrors(errors); err != nil {
+		return err
 	}
 
+	var errs []error
 	for _, repo := range templateRepos {
 		pullCmd := exec.Command("faas-cli", "template", "pull", repo)
 		pullCmd.Dir = filePath
-		err = pullCmd.Start()
-		if err != nil {
-			return fmt.Errorf("Failed to start faas-cli template pull: %t", err)
+		if err := pullCmd.Start(); err != nil {
+			errs = append(errs, fmt.Errorf("%s, %s", repo, err.Error()))
+			continue
 		}
-		err = pullCmd.Wait()
-		if err != nil {
-			return fmt.Errorf("Failed to wait faas-cli template pull: %t", err)
+		if err := pullCmd.Wait(); err != nil {
+			errs = append(errs, fmt.Errorf("%s, %s", repo, err.Error()))
 		}
 	}
 
+	err := joinErrors(errs)
+	return err
+}
+
+func joinErrors(errors []error) error {
+	if len(errors) > 0 {
+		var msg string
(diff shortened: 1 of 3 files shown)


git-tar/function/ops.go
```diff
@@ -41,32 +41,35 @@ type cfg struct {
...
 }
 
 func fetchTemplates(filePath string) error {
-	templateRepos := formatTemplateRepos()
+	templateRepos, err := formatTemplateRepos()
+	if err != nil {
+		return fmt.Errorf("Failed to format template repos: %t", err)
+	}
 
 	for _, repo := range templateRepos {
```
alexellis: We don't return where there are errors, but we used to. Why?

git-tar/function/ops.go
```diff
@@ -547,21 +550,22 @@ func importSecrets(pushEvent sdk.PushEvent, stack *stack.Services, clonePath str
...
+	var err error
 	if envTemplates := os.Getenv("custom_templates"); len(envTemplates) > 0 {
-		customTemplates := strings.Split(envTemplates, ",")
+		customTemplates := strings.Split(strings.Trim(envTemplates, " "), ",")
 		for _, repo := range customTemplates {
-			if strings.Contains(envTemplates, "https://") || strings.Contains(envTemplates, "http://") {
-				repo = strings.Trim(repo, " ")
-				templateRepos = append(templateRepos, repo)
+			repo = strings.Trim(repo, " ")
+			if _, err = url.ParseRequestURI(repo); err != nil {
```
alexellis: Looks like we could lose errors here because we don't return

git-tar/function/ops.go
Waterdrips: I was trying to cater for the scenario where a user has `n` invalid repos, 

If we returned on first error we would have to fix, and re-deploy `n` times. 

Ill re-work this to capture all errors and return a rollup

alexellis: I am happy with failing on the first one.

git-tar/function/ops.go
alexellis: What happened with this?

git-tar/function/ops.go
```diff
@@ -41,32 +41,48 @@ type cfg struct {
...
+		if err := pullCmd.Wait(); err != nil {
+			errs = append(errs, fmt.Errorf("%s, %s", repo, err.Error()))
 		}
 	}
 
+	err := checkErrorsList(errs)
+	return err
+}
+
+func checkErrorsList(errors []error) error {
```
alexellis: Rename this please, joinErrors()?

git-tar/function/ops.go
Waterdrips: Yeh, I moved it to accumulate a list of errors, then check the list a little later on - im not 100% sure im happy with the solution in the long run, but I couldn't think of a better solution at the time.

git-tar/function/ops.go
Waterdrips: done, plus tests added

PR #330, opened 2018-11-08
Fix dashboard link

ivanayov:
The stack-deploy link was trying to point to gateway_pretty_url, edited to a non-existing endpoint. It's now fixed to point to the dashboard URL

Signed-off-by: Ivana Yovcheva <iyovcheva@vmware.com>

## How Has This Been Tested?

Unit-tested with stack-deploy context.

Tested with `docwareiy/github-status:fix-dashboard-url-0.1`
`gateway_public_url: https://md-gw.team-serverless.xyz/`
generated system-dashboard link `https://system.md-gw.team-serverless.xyz/dashboard/ivanayov/`

## How are existing users impacted? What migration steps/scripts do we need?
Not impacted, just a link fix

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required) N/A
- [x] read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] signed-off my commits with `git commit -s`
- [x] added unit tests

PR #569, opened 2019-12-10
Fix dashboard link

Waterdrips:
## Description
On Github, when a user's stack-deploy event sucseeded the url pointed at their function domain, so in my case `waterdrips.o6s.io` for the community cluster, now it points to the dashboard for that user, so `system.o6s.io/dashboard/Waterdrips`

Also on failure this stack-deploy event link pointed to some non-existent location where a function would have been, based on the stack-deploy event name. Thus now also points to the user's dashboard.

Also, found an issue that when we renamed "log" page to "build-log" and "function-log" I missed updating this link from github-status, so it pointed at non-existent page :(

closes #330 

## How Has This Been Tested?

Unit tests, and deployed to my own OFC cluster and tests the stack-deploy success and failure scenarios, as well as failed and successful function builds (still point to correct places)

A new github-status image is needed and a user can update by editing the docker image of the github-status deployment.

The version is yet to be released but will probably be: `functions/git-status:0.3.7` - this image name is assumed for the rest of this section, if it is not correct then replace as appropriate

this can be updated by running: 

`kubectl set image deployment/github-status github-status=functions/github-status:0.3.7 --record -n openfaas-fn`

This is assuming the user is using a GitHub integration. Gitlab users need not change anything

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/github-status/handler.go
+++ b/github-status/handler.go
@@ -18,8 +18,6 @@ import (
 )
 
 const (
-	defaultPrivateKeyName   = "private-key"
-	defaultSecretMountPath  = "/var/openfaas/secrets"
 	githubCheckCompleted    = "completed"
 	githubCheckQueued       = "queued"
 	githubConclusionFailure = "failure"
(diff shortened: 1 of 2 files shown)


Waterdrips:
This is included in #571 and that has extra fixes too

PR #571, opened 2019-12-16
Fix url formatting when function exists in pretty url

### LINKED ISSUE #570: Self-Hosted installs with "function" in the domain name are reporting incorrect urls to user
A user has reported that a domain name with "function" in it reports some dashboard URLs incorrectly

Thinking about it, a sub-domain would also cause the issue

## Expected Behaviour

the dashboard URL and github status url would be correct

## Current Behaviour

So for a domain of:
 *.function.my-function-domain.com
*.my-function.com
wouldnt display the correct URLs in the dashboard (on function detail page) or on github-status page

## Possible Solution

the code does a "replace" on function, just the first occurrence (when a pretty_url is set)
There may be other instances of this in the codebase, so please look out for them!
 
1. set sub-domain on ofc to `function`
2. your dashboard will show incorrect url to invoke the function
3. github will also link to the wrong place

## Context

A user reported this on slack

Waterdrips:
## Description
On the dashboard, and on github status the pretty_url
formatting was replacing the first instance of function
in the pretty url, rather than the one at the end.

closes #570
closes #330 
closes #569 

## How Has This Been Tested?

deployed the dashboard to my ofc with a pretty_url of 
`https://user.function-heyal-user.uk/function`
my function name was `cat-api`
before the fixes the dashboard displayed: `https://waterdrips.cat-api-heyal-user.uk/cat-api`

After my change:
`https://waterdrips.function-heyal-user.uk/cat-api` as shown in screenshot
Screenshot 2019-12-16 at 15 14 45

The URL reported back to github was also wrong (see #330 #569 ) this is now correct
(tested by stack-deploy & a function success, and then a failure for both, point to correct places)

A new dashboard version is required for users who have "function" in their domain or sub-domain
A new github-status version is required for all users, this fixes the link as seen in #330 

## Checklist:

I have:

- [x] added unit tests


dashboard/client/src/api/functionsApi.test.js
```diff
@@ -70,4 +71,20 @@ describe('functionsApi', () => {
       expect(second.name).toEqual(functionResponse1.name);
     });
   });
+
+
+  describe('replaceFunctionEnding', () => {
```
alexellis: Please add a case where it does nothing, for when it's not appropriate - i.e. `function` is not in the sub-domain.

dashboard/client/src/api/functionsApi.js
```diff
@@ -52,7 +51,8 @@ class FunctionsApi {
       if (this.queryPrettyUrl) {
         endpoint = this.prettyDomain
           .replace('user', user)
-          .replace('function', shortName);
+        endpoint = replaceFunctionEnding(endpoint, shortName)
```
alexellis: `replaceFunctionEnding` <- this may be what the code does, but isn't it more like `formatFunctionURL` or `getFunctionURL`?

github-status/url_builder.go
```diff
@@ -0,0 +1,81 @@
...
+	if success && !isStack {
+		serviceValue := sdk.FormatServiceName(owner, service)
+		url = url + "function/" + serviceValue
+	} else {
+		url = url + "function/system-dashboard"
+	}
+	return url
+}
+
+func BuildPublicStatusURL(status, statusContext string, event *sdk.Event) string {
```
alexellis: Why is this method now public? 

What does the TODO comment mean?

Waterdrips: My bad on these, tried to get this in before leaving the office and missed that

dashboard/client/src/api/functionsApi.test.js
```diff
@@ -70,4 +71,26 @@ describe('functionsApi', () => {
...
+    it('replaces function with trailing slash', async () => {
```
alexellis: Can we have a test that does nothing, i.e. it's not required?

dashboard/client/src/api/functionsApi.js
```diff
@@ -1,11 +1,11 @@
 import axios from 'axios';
 import moment from 'moment';
+import {createNew} from "typescript";
```
alexellis: How come we're importing typescript into React?

Waterdrips: Think this was an IDE auto-import. removed

dashboard/client/src/api/functionsApi.test.js
Waterdrips: The function will always append the function name if the url doesn't end with `function`. Assuming that if a user has forgotten to put `/function` then they want the url to be `/<user-function-name>`

PR #576, opened 2020-01-04
Function logs when scaled to 0 gives warning

### LINKED ISSUE #575: Don't fetch logs for functions scaled to 0/N
Currently the page will show a loading spinner forever.

We should display a message or error

## Expected Behaviour

Spinner goes and we get a message
## Current Behaviour

spins forever
Screenshot_20200104-105236

## Possible Solution

set a failed flag, or something about it being scaled and display this in the UI

1. Wait for a fn to scale to 0
2. go check its logs

Waterdrips:
## Description

When we get an error fetching function logs then we now give a
warning that it could be due to functions being scaled to zero.

closes #575 

## How Has This Been Tested?

Deployed onto my OFC installation, waited for fn to scale to zero and tried to fetch logs:
Screenshot from 2020-01-04 15-45-40

then invoked the function and got logs again:

Screenshot from 2020-01-04 15-55-17

Upgrade of the OFC dashboard needed once it is released

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/dashboard/client/src/api/functionsApi.js
+++ b/dashboard/client/src/api/functionsApi.js
@@ -190,9 +190,13 @@ class FunctionsApi {
     const url = `${
         this.apiBaseUrl
     }/function-logs?function=${longFnName}&user=${user}`;
+
     return axios.get(url).then(res => {
-      return res.data;
-    });
+          return res.data;
+        }).catch(
+          fail => {
+            throw Error("Failed to fetch function logs - is the function scaled to 0? \nmessage:" + fail.message)
+        });
   }
 }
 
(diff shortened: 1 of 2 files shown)


alexellis:
In the error I'm seeing two sentences, but without any punctuation or separation between them. Could you look at this?

It might also be worth just stating: "can't show logs when the function is scaled to zero"

Waterdrips:
> In the error I'm seeing two sentences, but without any punctuation or separation between them. Could you look at this?
> 
> It might also be worth just stating: "can't show logs when the function is scaled to zero"

Changed to below:

Screenshot from 2020-01-06 08-57-14

alexellis:
Looks perfect.

PR #593, opened 2020-02-25
Lowercase the creation of ecr registries

Waterdrips:
#Description 

ECR call was failing for users with any uppercase in their usernames. I
had this... annoying.

## How Has This Been Tested?

This has been tested on a ECR enabled cluster, it creates the repo,
pushes images and pulls them correctly.

Users with Uppercase usernames can now push to ECR.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/register-image/handler.go
+++ b/register-image/handler.go
@@ -26,7 +26,7 @@ func (r ImageRequest) GetRepository() string {
 	if tagIndex > -1 {
 		image = image[:tagIndex]
 	}
-	return image
+	return strings.ToLower(image)
 }
 
 // Handle accepts an ImageRequest and creates the registry entry if required
--- a/register-image/handler_test.go
+++ b/register-image/handler_test.go
@@ -39,3 +39,15 @@ func Test_GetImageStripsECRRegistry(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func Test_GetImageLowecasesRegistryName(t *testing.T) {
+	r := ImageRequest{Image: "1238475.dkr.ecr.eu-central-1.amazonaws.com/ofbuilder/go-for-IT"}
+
+	got := r.GetRepository()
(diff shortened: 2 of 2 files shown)


alexellis:
My question is, why aren't we also running toLower upstream of this?

Waterdrips:
we build the image name in `git-tar` and pull the "owner" directly from the event, which comes 

```
imageName := formatImageShaTag(pushRepositoryURL, &v, pushEvent.AfterCommitID,
			pushEvent.Repository.Owner.Login, pushEvent.Repository.Name)
```

iv traced it back to `github-event` and it doesn't get to-lowered there, or on its way to `register-image`. It might be an idea to upstream lowering it? 

I suspect all the other places we are doing it at the end of the chain where the symptoms were

alexellis:
> iv traced it back to github-event and it doesn't get to-lowered there, or on its way to register-image. It might be an idea to upstream lowering it?

^ yes it would be worth doing.

Mixed-case usernames do work in the community-cluster, it may be addressed in those places already, but better to do it as high up as possible I would say, including a unit test for `formatImageShaTag`

Waterdrips:
> > iv traced it back to github-event and it doesn't get to-lowered there, or on its way to register-image. It might be an idea to upstream lowering it?
> 
> ^ yes it would be worth doing.
> 
> Mixed-case usernames do work in the community-cluster, it may be addressed in those places already, but better to do it as high up as possible I would say, including a unit test for `formatImageShaTag`

Ill raise an issue to track and do proper E2E testing if im changing stuff that high up in the chain.  Thanks for merging this.

alexellis:
`functions/register-image:0.2.0` created for this

Waterdrips:
> `functions/register-image:0.2.0` created for this

Thanks, Iv deployed this image and tested :taco: :tada: all looks good, Thanks!!

PR #599, opened 2020-03-03
Update dashboard to use node12 template

alexellis:
Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

## Description

Update dashboard to use node12 template

node10 is out of LTS and so node12 should be used, the template
works differently though - with async/await vs callbacks. This
meant moving the fs.readFile over to promises and the requestsjs
module to axios.

## How Has This Been Tested?

Testing:
- [x] Test locally without OAuth
- [x] Test on the community cluster with auth + TLS enabled

No change other than updating the version

### DIFF
--- a/dashboard/stack.yml
+++ b/dashboard/stack.yml
@@ -4,16 +4,11 @@ provider:
 
 functions:
   system-dashboard:
-    lang: node10-express
+    lang: node12
     handler: ./of-cloud-dashboard
-    image: functions/of-cloud-dashboard:0.7.3
+    image: alexellis2/of-cloud-dashboard:0.7.17
     labels:
       openfaas-cloud: "1"
       role: openfaas-system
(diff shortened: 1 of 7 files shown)


dashboard/stack.yml
```diff
@@ -4,16 +4,11 @@ provider:
 
 functions:
   system-dashboard:
-    lang: node10-express
+    lang: node12
     handler: ./of-cloud-dashboard
-    image: functions/of-cloud-dashboard:0.7.3
+    image: alexellis2/of-cloud-dashboard:0.7.18
```
Waterdrips: this should be a functions/ image when merged

alexellis:
@Waterdrips FYI the image is - functions/of-cloud-dashboard:0.8.0

PR #602, opened 2020-03-04
[Feature request] Support Github Orgs as Customers in edge-auth

Waterdrips:
## Description
This Commit adds support for using a github Organisation as a "Customer"
in the customers file. This means users can manage their access controls
using github organisations rather than having to add/remove from the
customers file.

## How Has This Been Tested?

Tested by only having an organisation in the Customers file, checking
that I didnt get access with the previous version of edge-auth, then
changing to use my new version and I was issued a token and could access
the dashboard

Unit tests written

Any users who have an organisation in their Customers file will give access to public members of that organisation if they don't already exist in the ACL.

Users who wish to manage their ACL using github organisations can now use the organisation in the Customers file. Then only need to add/remove users from the organisation to give access.

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -142,7 +142,8 @@ private_key_filename: my-private-key
 
 Edit `customers_url` in gateway_config.yml.
 
-Enter a list of GitHub usernames for your customers, these are case-sensitive.
+Enter a list of GitHub usernames for your customers, these are case-sensitive. You can use Organisations in place of 
+usernames, anyone that is a public member of the organisation can then login.
 
 ### Customize for Kubernetes or Swarm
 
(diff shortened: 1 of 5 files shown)


alexellis:
Hi @Waterdrips, changes like this which affect the roadmap and security need to be properly proposed as issues before being worked on.

See the contribution guide about making proposals, please include the use-case, current workarounds and why you feel this change is required.

https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#i-have-a-great-idea

Alex

Waterdrips:
/close 

see #603

PR #608, opened 2020-03-06
Allow the use of Secure cookie for edge-auth

### LINKED ISSUE #605: Set edge-auth's cookie to use "Secure" (HTTPS only)
Set edge-auth's cookie to use "Secure" (HTTPS only)

## Expected Behaviour

The most secure configuration would be for us to set the Secure: true flag for the cookie in edge-auth, this prevents the cookie from being used over a plaintext (HTTP) connection.

The downsides here are that nobody can test without setting up full TLS i.e. LetsEncrypt, which tends to be a pain for development.

I would like to see this behaviour behind a flag, so that folks can still work locally on OAuth features without LetsEncrypt.

## Current Behaviour

The HTTP Only variable means that the token can't be read from JavaScript, and we use HTTPS on customer and public environments, so this change won't make much difference to existing users, it's more belt-and-braces.

## Possible Solution

Update this code https://github.com/openfaas/openfaas-cloud/blob/6df6e79f2b69c0f6b57104f4b75d4dce5d8a0701/edge-auth/handlers/oauth2.go#L154

**But read the setting from the edge-auth configuration.**

Waterdrips:
## Description
The "Secure" setting for cookies was not able to be set, we have enabled
the parameter and set it to default to `false`, maintaining existing
functionality.

Optionally a user can now set "secure_cookie" to true in the edge-auth
deployment and this will set the flag on the user's cookie, meaning it
will only be transmitted over HTTPS.

I have updated the Helm Chart to include this option. The images have not been bumped in the deployments as they have not been created yet. 

Fixes #605 

## How Has This Been Tested?

Logged into working installation.
Deleted Cookie
Updated a deployment with the new image, no secure_cookie env var set on the deployment:
got a cookie, and it was secure=false

Deleted cookie
Updated the same deployment to add the secure_cookie=true env var.
Authd with the OFC install, got a cookie and it had "secure=true" set.

Delete cookie
Updated the same deployment to change the secure_cookie to false.
Authd, got a cookie and it had "secure=false" set.

A user with an existing instillation, if they updated the image without adding the new env var they would see no change, and it would work.

A user who wanted to use TLS (recommended for production) could ensure that their cookies were only send over HTTPS. 

## Checklist:

I have:

- [x] added unit tests

### DIFF
--- a/edge-auth/handlers/config.go
+++ b/edge-auth/handlers/config.go
@@ -14,6 +14,7 @@ type Config struct {
 	Scope                  string
 	CookieRootDomain       string
 	CookieExpiresIn        time.Duration
+	SecureCookie           bool
 	PublicKeyPath          string
 	PrivateKeyPath         string
 	Debug                  bool // Debug enables verbose logging of claims / cookies
--- a/edge-auth/main.go
+++ b/edge-auth/main.go
@@ -30,6 +30,7 @@ func main() {
 	var oauthClientSecretPath string
 
 	var writeDebug bool
+	secureCookie := false
 
 	if val, exists := os.LookupEnv("oauth_provider"); exists {
 		oauthProvider = val
@@ -63,6 +64,13 @@ func main() {
 		cookieRootDomain = val
 	}
 
+	if val, exists := os.LookupEnv("secure_cookie"); exists {
+		if boolVal, err := strconv.ParseBool(val); err == nil {
+			secureCookie = boolVal
+		}
+		log.Printf("unable to convert %s to bool for %q", val, "secure_cookie")
+	}
+
 	if val, exists := os.LookupEnv("public_key_path"); exists {
 		publicKeyPath = val
 	}
@@ -88,6 +96,7 @@ func main() {
 		CookieRootDomain:       cookieRootDomain,
 		ExternalRedirectDomain: externalRedirectDomain,
 		Scope:                  "read:org,read:user,user:email",
+		SecureCookie:           secureCookie,
 		PublicKeyPath:          publicKeyPath,
 		PrivateKeyPath:         privateKeyPath,
 		OAuthClientSecretPath:  oauthClientSecretPath,
--- a/yaml/core/edge-auth-dep.yml
+++ b/yaml/core/edge-auth-dep.yml
@@ -73,6 +73,10 @@ spec:
           - name: cookie_root_domain
             value: ".system.o6s.io"
 
+# Config for setting the cookie to "secure", set this to true for HTTPS only OAuth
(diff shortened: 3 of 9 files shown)


chart/openfaas-cloud/values.yaml
```diff
@@ -22,6 +22,7 @@ edgeAuth:
   writeDebug: false
   ## clientSecret is the value from your OAuth application
   clientSecret:
+  secureCookie: "false"
```
alexellis: Should we default to true?

edge-auth/handlers/config.go
```diff
@@ -14,6 +14,7 @@ type Config struct {
 	Scope                  string
 	CookieRootDomain       string
 	CookieExpiresIn        time.Duration
+	SecureCookie		   bool
```
alexellis: Can you run `gofmt -s -w` on this file please?

edge-auth/main.go
```diff
@@ -30,6 +30,7 @@ func main() {
 	var oauthClientSecretPath string
 
 	var writeDebug bool
+	secureCookie := false
```
alexellis: Default to true, unless set to false.

edge-auth/main.go
```diff
@@ -88,6 +96,7 @@ func main() {
 		CookieRootDomain:       cookieRootDomain,
 		ExternalRedirectDomain: externalRedirectDomain,
 		Scope:                  "read:org,read:user,user:email",
+		SecureCookie:			secureCookie,
```
alexellis: `gofmt` needed

yaml/core/edge-auth-dep.yml
```diff
@@ -73,6 +73,10 @@ spec:
           - name: cookie_root_domain
             value: ".system.o6s.io"
 
+# Config for setting the cookie to "secure", set this to true for HTTPS only OAuth
+          - name: secure_cookie
```
alexellis: Default to true

chart/openfaas-cloud/values.yaml
Waterdrips: the default tls.enabled value is "False" we could use the same config value to work out if we should enable it rather than bringing in a new one?

edge-auth/main.go
Waterdrips: This was left default false so a user upgrading their OFC installation by just bumping the img tag would get the same functionality. 

This is because the env var may not be set/present?

alexellis:
Hi @Waterdrips thank you for this PR 💪 

Tag `0.7.3` has been built and released if you'd like to update the YAML files.

PR #611, opened 2020-03-06
Make Github Status namespace checks by APP ID

### LINKED ISSUE #540: [Enhancement] Support multiple OFC apps per repository
## Expected Behaviour

I would like to be able to have a repository built and deployed onto multiple OFC installations and see the build status set successfully for each app on each commit.

## Current Behaviour

The integration with multiple OFC installations is not great - It works but does not report back that the deployment was successful and you only see each "build" step in 1 app. (screenshot)
image

## Possible Solution

namespace the build steps to the cluster or something so they can appear under both apps

1. Add multiple OFC integrations to a repository
2. Trigger a build

## Context

commits always sit in pending status and dont succeed.

## Your Environment

OFC Community and OFC personal, through Github integration

* Please also follow the [troubleshooting guide](https://github.com/openfaas/faas/blob/master/guide/troubleshooting.md) and paste in any other diagnostic information you have:

Waterdrips:
## Description

The Github Status function was not namespacing the check-runs. It was
using the same ting for each installed app. This has been changed to
namespace by app_id. This means that 2 + installations on the same
repository can now correctly create & report github check runs.

Fixes #540 

## How Has This Been Tested?

This has been tested by installing 2x OCF installations, then running a
pipeline. This failed of half of the functions as they didn't get the
correct check for that app.

Then I added the new function to each OFC installation, and re-ran a
pipeline. Every check run was created and reported back correctly.

Users with more than 1 app installed can correctly get statuses back from the OFC installations.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/github-status/handler.go
+++ b/github-status/handler.go
@@ -119,13 +119,15 @@ func getLogs(status *sdk.CommitStatus, event *sdk.Event) (string, error) {
 }
 
 func reportToGithub(commitStatus *sdk.CommitStatus, event *sdk.Event) error {
+	appID := os.Getenv("github_app_id")
 	if os.Getenv("use_checks") == "false" {
-		return reportStatus(commitStatus.Status, commitStatus.Description, commitStatus.Context, event)
+		return reportStatus(commitStatus.Status, commitStatus.Description, appID, event)
 	}
 	return reportCheck(commitStatus, event)
 }
 
 func reportStatus(status string, desc string, statusContext string, event *sdk.Event) error {
+	appID := os.Getenv("github_app_id")
 
 	ctx := context.Background()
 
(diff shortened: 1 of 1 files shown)


github-status/handler.go
```diff
@@ -119,21 +119,23 @@ func getLogs(status *sdk.CommitStatus, event *sdk.Event) (string, error) {
 }
 
 func reportToGithub(commitStatus *sdk.CommitStatus, event *sdk.Event) error {
+	appID := os.Getenv("github_app_id")
 	if os.Getenv("use_checks") == "false" {
-		return reportStatus(commitStatus.Status, commitStatus.Description, commitStatus.Context, event)
```
alexellis, line 123: What kind of value was in `commitStatus.Context` before vs the static GitHub app ID?

Waterdrips: https://developer.github.com/v3/repos/statuses/

```
context | string | A string label to differentiate this status from the status of other systems. Default: default
```

`default` was always the value

alexellis: OK, so did you pick the GitHub app ID because it should be unique between installations? It could be any identifier as long as it is consistent?

Waterdrips: Yeh, and its already included in the image as an env var

Waterdrips:
@alexellis any thoughs on merging this?

PR #612, opened 2020-03-07
Don't error if ECR Repository exists already

### LINKED ISSUE #539: False error from register-image when image is present on ECR already
- The ECR implementation gives errors when the registry already exists (at least git-tar and register-image gives exit(1) and status code != 200)
- The name that `register_image` is using could have potential problem, today the name is `registry_url`/`github-org´-`function-name?`-`image`:`image-tag`-`repo-branch`-`repo-sha` on selected region. The problem is that when you have multiple clusters that have deployed the same stack you will suddenly start to share registry in pipelines. It could create some strange edge cases
- I think `register_image` should add a basic policy when the registry is created so we dont need to go back to fix this through other jobs

1.
2.
3.
4.

## Your Environment

* ofc-bootstrap
* AWS EKS

Waterdrips:
## Description
We were returning an error if the ECR repo already exists. This shouldnt
give an error because it's not a failure. If the repo doesn't exist then
we create it, if it exists then that's fine.

Fixes #539 

## How Has This Been Tested?

Tested by deploying an ECR backed OFC installation, first time round it
creates the new repo, then after that it's not throwing an error, only
saying it already exists.

After - when trying to create repo that exists
```sh
2020/03/07 19:32:17 Attempting to create repo: waterdrips-openfaas-cloud-functions-roll
Repo Exists: waterdrips-openfaas-cloud-functions-roll

```

New deployment of register-image required. This fixes some output not to return an error.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/register-image/Gopkg.lock
+++ b/register-image/Gopkg.lock
@@ -80,6 +80,7 @@
   analyzer-version = 1
   input-imports = [
     "github.com/aws/aws-sdk-go/aws",
+    "github.com/aws/aws-sdk-go/aws/awserr",
     "github.com/aws/aws-sdk-go/aws/session",
     "github.com/aws/aws-sdk-go/service/ecr",
     "github.com/openfaas/openfaas-cloud/sdk",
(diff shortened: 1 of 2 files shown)


alexellis:
Thanks for the fix.

PR #625, opened 2020-04-07
Add marceliniho to CUSTOMERS file

marceliniho:
Add marceliniho to CUSTOMERS file

Signed-off-by: Marcel Adam <marcel.adam@experteach.de>

### DIFF
--- a/CUSTOMERS
+++ b/CUSTOMERS
@@ -158,3 +158,4 @@ transhapHigsn
 lukwil
 pbaezab
 telarpress
+marceliniho


marceliniho:
I have just created a training course for serverless computing and have therefore dealt with various solutions. Furthermore, we develop applications internally, which we want to switch to serverless platforms. These applications are currently web applications, which have to process many parallel jobs. We want to be more agile, so such a platform would be of great help. As a further point, I would like to show such of solutions in trainings.

PR #633, opened 2020-04-26
Add Cookie root domain env variable to Dashboard

### LINKED ISSUE #632: Bug - Logout button not removing a user's cookie
pressing logout on the dashboard should be setting the user's cookie for oauth to "" (empty string)

There is an issue in the deployment files because "cookie_root_domain" is not being set in the dashboard deployment. 

## Expected Behaviour

Logout clears a users cookie

## Current Behaviour

The cookie is not removed from the user's browser session

## Possible Solution

The "cookie_root_domain" env var isnt being set on the dashboard deployment, so we try and set a cookie for the domain "undefined", never removing the user's cookie. The workaround is editing the deployment and setting an env var of "cookie_root_domain" to ".system.your-ofc-domain"

ofc dployment files and ofc-bootstrap need this value adding to the dashboard deployment.

1. Click logout on the dashboard when using oauth

## Context

The "cookie_root_domain" env var isnt being set on the dashboard deployment, so we try and set a cookie for the domain "undefined", never removing the user's cookie. The workaround is editing the deployment and setting an env var of "cookie_root_domain" to ".system.your-ofc-domain"

Waterdrips:
## Description

Somewhere along the line the cookie_root_domain was missed from the
dashboard config. This only needs to be set when using OAuth, to remove
a users token from their cookie. This was not set and therefore caused
us to try and set the cookie domain to "undefined"

Fixes #632 

## How Has This Been Tested?

This has been tested by adding the env to the Dashboard deployment and
hitting logout on an OAuth enabled installation. This now sets the
correct cookie domain and removes the existing token

This is a bugfix, users will need to edit their dashboard deployment to add this environment variable

```
kubectl edit deploy -n openfaas-fn system-dashboard
```
and add the following to the environment variables section:
(example where your openfaas root domain is "example.com"
```
        - name: cookie_root_domain
          value: .system.example.com
```

## Checklist:

I have:

- [ ] added unit tests

This will also need amending in ofc-bootstrap to propagate this change

### DIFF
--- a/dashboard/dashboard_config.yml
+++ b/dashboard/dashboard_config.yml
@@ -10,3 +10,5 @@ environment:
   pretty_url: https://user.o6s.io/function
   # Comment out if not using public pretty-URL
   query_pretty_url: 'true'
+  # Cookie root domain, set is using OAuth, used to remove a user's cookie
+  cookie_root_domain: 'system.o6s.io'
--- a/dashboard/README.md
+++ b/dashboard/README.md
@@ -81,7 +81,7 @@ Set `public_url` to be the URL for the IP / DNS of the OpenFaaS Cloud.
 
 Set `cookie_root_domain` when using auth.
 
-Example with domain `o6s.io`:
+Example with domain `.system.o6s.io`:
 
 ```
 cookie_root_domain: ".system.o6s.io"


dashboard/dashboard_config.yml
```diff
@@ -9,4 +9,6 @@ environment:
   # Comment out if not using public pretty-URL
   pretty_url: https://user.o6s.io/function
   # Comment out if not using public pretty-URL
-  query_pretty_url: 'true'
\ No newline at end of file
+  query_pretty_url: 'true'
+  # Cookie root domain, set is using OAuth, used to remove a user's cookie
+  cookie_root_domain: 'system.o6s.io'
```
alexellis, line 14: This is templated, so wouldn't this need to be changed in ofc-bootstrap rather than here?

alexellis:
Happy to merge, but wouldn't this need to be changed in ofc-bootstrap?

dashboard/dashboard_config.yml
alexellis: This needs to be synced here -> https://github.com/openfaas-incubator/ofc-bootstrap/blob/master/templates/dashboard_config.yml

Waterdrips: Yeh, it will need changing there. The ones in that yaml looked like "example" settings so I added there for future reference.

PR #674, opened 2020-10-13
Remove of-client-secret as env var from edge-auth

### LINKED ISSUE #673: Deprecate passing Oauth client secret in environment for OFC chart
## Expected Behaviour

To pass the of-client secret only using a mounted secret

## Current Behaviour

it can be passed in env and secret. This should probably be standardised in one, as the only consumer of the chart at the moment is ofc-bootstrap and that's using the secret, it doesn't make sense to keep both.

## Possible Solution

remove the env variable option

Waterdrips:
## Description

Edge-auth now only uses the mounted secret to read the of-client-secret
for Oauth2. This is to reduce complexity in the chart and deployments.

Fixes #673 

## How Has This Been Tested?

the golang chart tests pass. This functionality has been in use for a long time, iv not seen the env variable being used over the secret anywhere.

## Checklist:

I have:

- [ ] added unit tests


alexellis (approved):
Approved

wilsonianb:
There's a couple stray `clientSecret`s that are causing an error when building the `edge-auth` image.
This fixed it for me https://github.com/wilsonianb/openfaas-cloud/commit/f02c4f69f854fdd9c880ddd28f4883d3e0677d6b

Waterdrips:
apologies - fixed now based on your commit :+1:

PR #678, opened 2020-10-14
Remove redundant setting for scheme in chart

### LINKED ISSUE #677: Remove redundant `global.scheme` from ofc helm chart
The chart can derive the value of this from `tls.enabled` so it's not required.

We can remove it from the chart and README for chart.

Waterdrips:
The chart had a setting that can be derived based on another value. This
commit removed the extra config value as it is not required

## Description
Fixes #677 

## How Has This Been Tested?

There's unit tests

N/A

## Checklist:

I have:

- [ ] added unit tests


alexellis (approved):
Approved

PR #680, opened 2020-10-15
Update chart namespaces to match existing

Waterdrips:
## Description
The namespaces in the chart were wrong or missing for a lot of
the resources. These have now been corrected along with adding the
annotation for ingress shim for cert-manager to get tls for domains if
they are asked for

## How Has This Been Tested?

Unit tests

### DIFF
--- a/chart/openfaas-cloud/templates/tls/issuer-prod.yml
+++ b/chart/openfaas-cloud/templates/tls/issuer-prod.yml
@@ -1,6 +1,6 @@
 {{- if .Values.tls.enabled }}
 apiVersion: cert-manager.io/v1alpha2
-kind: ClusterIssuer
+kind: Issuer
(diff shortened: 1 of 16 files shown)


chart/openfaas-cloud/templates/tls/issuer-prod.yml
```diff
@@ -3,7 +3,7 @@ apiVersion: cert-manager.io/v1alpha2
 kind: ClusterIssuer
```
alexellis, line 3: Should we move to a namespace-based issuer?

And keep `  namespace: {{ .Values.global.coreNamespace }}		`

Waterdrips: sure

PR #684, opened 2020-11-07
Migrate Travis CI to Github Actions

### LINKED ISSUE #683: Migrate from Travis to Github Actions
As proposed by @alexellis we should move all our repos to Github Actions because of it's generous free-tier as we're not able to get enough builds out of the free credits that Travis provides.
## Current Behaviour

We're running all our CI builds using Travis CI
## Possible Solution

Move to Github Actions

nitishkumar71:
## Description
We want to migrate from Travis CI to Github Actions
Fixes #683 
openfaas/faas#1585
## How Has This Been Tested?

Please check [CI Only](https://github.com/nitishkumar71/openfaas-cloud/actions/runs/351165720) Build
Please check [Publish](https://github.com/nitishkumar71/openfaas-cloud/actions/runs/355160257) build for tag push

1. We are not able to run CI Pipelines
2. We are moving to github actions for CI Pipelines

## Checklist:

I have:

- [ ] added unit tests


.github/workflows/ci-only.yml
```diff
@@ -0,0 +1,32 @@
...
+  push:
+    branches: [ "*" ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        go-version: [1.9.x]
```
alexellis, line 13: This should be 1.13

.github/workflows/ci-only.yml
```diff
@@ -0,0 +1,32 @@
+name: ci-only
+
+on:
+  push:
+    branches: [ "*" ]
+  pull_request:
+    branches: [ master ]
```
alexellis, line 7: Why only master?

alexellis (changes requested):
I have some suggestions

.github/workflows/ci-only.yml
nitishkumar71: So, should we enable it for both push and pull request?

alexellis:
@wilsonianb could you advise re: pushing tags based upon the release version?

.github/workflows/ci-only.yml
nitishkumar71: Done

wilsonianb:
You could do something like this (in an `on.tag` workflow) for the non-function components and use `${{ github.ref }}` for the image tags:
https://github.com/wilsonianb/openfaas-cloud/blob/dfc967a49eca9acd099afc02138d96607ea4332b/.github/workflows/docker.yml#L34-L57

nitishkumar71:
> You could do something like this (in an `on.tag` workflow) for the non-function components and use `${{ github.ref }}` for the image tags:
> https://github.com/wilsonianb/openfaas-cloud/blob/dfc967a49eca9acd099afc02138d96607ea4332b/.github/workflows/docker.yml#L34-L57

@wilsonianb Thank you sharing the example.

@alexellis Can you please review it.

Waterdrips:
@nitishkumar71 There's actually a lot more to converting ofc into actions than adding these workflows. As OFC is a complex system with lots of moving parts its probably best for me to take this over.

nitishkumar71:
> @nitishkumar71 There's actually a lot more to converting ofc into actions than adding these workflows. As OFC is a complex system with lots of moving parts its probably best for me to take this over.

makes sense

PR #685, opened 2020-11-08
Remove a print of customers in edge-auth

Waterdrips:
## Description
This removes a printing of the customers in edge-auth.

reduced log noise

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/edge-auth/handlers/query.go
+++ b/edge-auth/handlers/query.go
@@ -30,7 +30,13 @@ func MakeQueryHandler(config *Config, protected []string, restrictedPrefix []str
 	customersURL := os.Getenv("customers_url")
 
 	customers := sdk.NewCustomers(customersPath, customersURL)
-	customers.Fetch()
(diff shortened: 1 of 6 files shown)


edge-auth/handlers/query.go
```diff
@@ -30,7 +30,10 @@ func MakeQueryHandler(config *Config, protected []string, restrictedPrefix []str
 	customersURL := os.Getenv("customers_url")
 
 	customers := sdk.NewCustomers(customersPath, customersURL)
-	customers.Fetch()
+	if err := customers.Fetch(); err != nil {
+		log.Fatalf("Unable to fetch customers from configured source")
```
alexellis, line 34: Not sure if this should be fatal. What if it timed out?

Waterdrips: then the next time they try it might not? changed the log line and written a 500 internal server error instead.

alexellis (approved):
Approved

PR #686, opened 2020-11-10
Part one of converting OFC to go modules

Waterdrips:
## Description

This will initialise a module in each function/folder. This needs to be
merged before modules can start to manage dependancies because otherwise
you get a duplicate definition error as each function that imports
anything from the ofc repo (from github in its gopkg) will pull the
whole repo in, not just the module it wants (the SDK for example) so we
need a go mod in each folder to avoid it existing twice in the
dependency tree for the function.

Once this is in, each module can be converted to go modules and remove
the use of dep

## How Has This Been Tested?

I added a replace directive for the local module on disk and tested the go mod vendor and go mod tidy calls worked and started to manage deps. Without the replace we got the errors mentioned above

no changes. This is background work to upgrade to go modules and is required to make the switch.

## Checklist:

I have:

- [ ] added unit tests

### DIFF
--- a/audit-event/go.mod
+++ b/audit-event/go.mod
@@ -1,0 +1,11 @@
+module github.com/openfaas/openfaas-cloud/audit-event
+
+go 1.13
+
+require (
+	github.com/alexellis/hmac v0.0.0-20180624210714-d5d71edd7bc7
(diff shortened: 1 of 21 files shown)


Waterdrips:
first part of https://github.com/openfaas/openfaas-cloud/issues/687

alexellis (approved):
Approved

PR #689, opened 2020-11-12
Github Actions migration

Waterdrips:
## Description
Big PR to move over to GH Actions for building and publishing OFC containers

This will build containers and stack.yml and publish new versions with a tag that matches a release tag.

We will need to switch the chart to be using the release tag and pull from ghcr.io rather than dockerhub next.

Closes #684 

## How Has This Been Tested?

Containers are built and publised on relese tag
They are built and the chart tests are run on commits and PRs

We will be moving to GHCR for hosting images so will need to update the chart once they are pushed and public.
We will then need to update the chart to use the release tag as the image versions as we are going to be pushing those image tags from now on, rather than keeping each container on its own version. 

This should make maintaining the chart and stack.yml easier going forward. It will however mean that users should be updating all of ofc components on each release rather than just the components that change. This should be easier once the helm chart goes into ofc-bootstrap as theres a flag being added to run only the chart install

## Checklist:

I have:

- [ ] added unit tests


Waterdrips:
https://github.com/openfaas/faas/issues/1585 part of this

.github/workflows/build.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          fetch-depth: 1
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Run CI
+        run: sh ./contrib/ci.sh
+
+  # Test building the non-cli Containers
```
alexellis, line 47: Non-cli containers?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          fetch-depth: 1
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Login to Docker Registry
+        uses: docker/login-action@v1
+        with:
+          username: ${{ github.repository_owner }}
```
alexellis, line 28: Wait, who is the owner when it's an org? Can you test and find out please?

alexellis: If you use DOCKER_USERNAME it will just be easier

Waterdrips: secrets.DOCKER_USERNAME will cause every occurrence of the username to be starred out, which is annoying.

alexellis (approved):
Approved

.github/workflows/publish.yaml
alexellis: We have that everywhere else and I don't think that openfaas can push images? I can't get a PTA for openfaas the owner.

alexellis:
This looks great - we are getting there!

I'm not seeing the gitlab stack file being built? How about the aws one too?

